Issue #92, opened 2019-02-01
Consider renaming `tp-ui-core-font-face` to `thumbprint-font-face`


tomgenoni:
This would align it with the other packages. The `tp-ui...` syntax is not needed once #90 and #91 are complete.

danoc:
Agreed. 👍

PR #96, opened 2019-02-02
Thumbprint SCSS package

tomgenoni:
Items to address:

- [x] I moved the `README.mdx` files into a `docs` dir, move them as needed
- [x] Since we're just copying the root mixin SCSS file and the mixins there's no compiling. There is no output of `thumbprint-mixin.scss` but it's possible a future change would break a mixin path. How to test? I suppose we could compile it to an empty file and throw it away.
- [x] Replace references to `tp-ui-*` in this codebase with the new references. The only exception is `tp-ui-core-font-face` until #92 is done.


danoc:
@tomgenoni — Docs are all working 👍

danoc:
@tomgenoni — I've locked the deploy. Feel free to merge.

PR #97, opened 2019-02-04
Add info about how we add Atomic classes

### LINKED ISSUE #18: Add docs that explain when we add classes to Atomic
People often ask us to add classes to Atomic. It'd be great to have text in the docs that explains how we decide what to add to Atomic.

Past discussion:
* https://github.com/thumbtack/thumbprint-archive/issues/1266
* https://github.com/thumbtack/thumbprint-archive/issues/1355

tomgenoni:
Fixes #18

### DIFF
--- a/www/src/pages/atomic/usage.mdx
+++ b/www/src/pages/atomic/usage.mdx
@@ -102,7 +102,7 @@ As with the previous Sass example above, since this CSS is not included in our c
 
 ## Adding new Atomic classes
 
-We occasionally get requests to add new classes to the Atomic library. In an effort to balance the library’s completeness with its bundle size, we’ve adopted the approach that there should be at least 10 instances of a property before we add it to Atomic.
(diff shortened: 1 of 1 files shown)


www/src/pages/atomic/usage.mdx
```diff
@@ -100,6 +100,12 @@ As with the previous Sass example above, since this CSS is not included in our c
 }
 ```
 
+## Adding new Atomic classes
+
+We occasionally get requests to add new classes to the Atomic library. In an effort to balance the library’s completeness with its bundle size, we’ve adopted the approach that there should be at least 10 instances of a property before we add it to Atomic.
```
danoc:
```suggestion
We occasionally get requests to add new classes to the Atomic library. In an effort to balance the library’s completeness with its bundle size, we’ve adopted the approach that there should be at least 10 instances of a property in Thumbtack's website codebase before we add it to Atomic.
```

PR #109, opened 2019-02-08
Add docs for new packages

tomgenoni:
There's one bug where 'Globals' is showing up both at the top where expected and in the general list due to the logic that generates the list.


www/src/pages/overview/developers.mdx
```diff
@@ -0,0 +1,39 @@
...
+## React projects
+
+### RATS
+
+No, not the little animal. It’s an acronym to help remember the order of technologies to try.
+
+1. **R**eact
+2. **A**tomic
+3. **T**okens
+4. **S**ass
```
danoc: Not sure if I understand this one.

Does it mean custom CSS in the developer's own project? Or our Thumbprint SCSS components?

If it's custom CSS in the developer's project, then this could probably merged into the description for the "Tokens" section since tokens are almost always used in custom CSS files.

tomgenoni: I do see them as separate topics though maybe it's not clear enough. 

Tokens should be separate because we make them available in JS they are used occasionally:

``
import {
    tpColorBlue,
    tpColorBlack300,
    ...
}
```

By "custom CSS", which is what I mean by the fourth item, I mean token-less Sass. I can drop the acronym and make it clearer that...

Tokens = JS tokens in JSX and Sass tokens in Sass modules
Custom Sass/CSS =  hardcoded values where no token is appropriate

danoc: > By "custom CSS", which is what I mean by the fourth item, I mean token-less Sass. I can drop the acronym and make it clearer that...

Ah, ok. Yeah, I think that can be clarified. I thought it meant Sass with tokens.

tomgenoni: Updated to be clearer.

Issue #24, opened 2018-11-05
Documentation is not loading in IE 11


danoc:
I think this is caused by `gatsby-mdx`. Workaround is for me or @tomgenoni to publish beta packages.

PR #117, opened 2019-02-12
Improve IE 11 support

danoc:
This helps fix an issue where pages are not loading in IE 11.


danoc:
Relates to #24.

danoc:
@jichan-thumbtack — I've been digging into this for a while and found some more ways to improve it. There's still an issue with Gatsby MDX that I'll continue to track down.

PR #124, opened 2019-02-15
Fix file watching/reloading for components

danoc:
This updates Gatsby's webpack config to prefer the local copy of our components when developing locally.

### DIFF
--- a/www/gatsby-node.js
+++ b/www/gatsby-node.js
@@ -22,10 +22,12 @@ exports.onCreateWebpackConfig = ({ actions }) => {
         resolve: {
             modules: [path.resolve(__dirname, 'src'), 'node_modules'],
             alias: {
-                // Use Thumbprint React and SCSS source files when `import`'ing them in the
-                // documentation. By default, Webpack will use the value of the `main` or `module`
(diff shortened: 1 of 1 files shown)


www/gatsby-node.js
```diff
@@ -21,6 +21,22 @@ exports.onCreateWebpackConfig = ({ actions }) => {
     actions.setWebpackConfig({
         resolve: {
             modules: [path.resolve(__dirname, 'src'), 'node_modules'],
+            alias: {
+                // Use Thumbprint React and SCSS source files when `import`'ing
+                // them in the documentation. By default, Webpack will use the
+                // value of the `main` or `module` field in the package's
+                // `package.json` file. This makes it difficult to develop
+                // locally because it'd require running re-compiling the code
```
tomgenoni: "because it'd require running re-compiling the code" do you mean "it would require re-compiling..." ?

danoc: Yup, good catch. Did the comment make sense otherwise?

tomgenoni: Yup

PR #129, opened 2019-02-19
Use custom `LazyImage` component instead of lazysizes.

### LINKED ISSUE #69: Remove `lazysizes` usage
Developers are getting this error in `website` since it forces a specific version of `lazysizes` that is different than Thumbprint.

```
├─┬ @thumbtack/thumbprint-react@0.1.1
│ └── UNMET DEPENDENCY lazysizes@^4.1.5
└── lazysizes@4.0.4

npm ERR! missing: lazysizes@^4.1.5, required by @thumbtack/thumbprint-react@0.1.1
```

We should:

1. Understand why `website` forces a specific version. (Perhaps it's unsafe to load twice on a page?)
2. Consider downgrading the Thumbprint version, if needed. Or upgrading the `website` version.

***
Thanks @xanderwebs for the report!

danoc:
This creates a _internal-only_ `Image` component in Thumbprint that is used by `Avatar` and `ServiceCard`.

It is the first step in moving away from `lazysizes` in favor of components that works better with React.

The `LazyImage` component in this CR is exactly the same as the component used to lazy-load Pro photos on pages such as: https://www.thumbtack.com/ca/san-francisco/house-cleaning/

**Why keep this component private?** We're still working out an API that will make it easy for developers to use responsive images (and images from the image service). We may make it public and document once we're confident about the API.

***
**Previews:**
https://deploy-preview-129--thumbprint.netlify.com/components/avatar/react/
https://deploy-preview-129--thumbprint.netlify.com/components/service-card/react/
***

Fixes #69.


www/src/components/mdx/code-block/styles.js
```diff
@@ -8,7 +8,7 @@ const previewThemes = {
 
 const classes = {
     // Border color is defined by the theme above.
-    preview: `pa4 mb1 ba bw-2 br2 overflow-x-auto ${styles.preview}`,
```
danoc: This change removes the `x` overflow from the React live examples. It's needed because `react-waypoint`, the tool used to determine if an element is visible, used the element's first scrollable parent as the reference. This means that if the element is visible within the scrollable, then `react-waypoint` runs the `onEnter` function.

This default behavior is desired in most cases, but caused problems in our case because the elements were _always_ visible within the live previews.

We could provide a prop to control [React Waypoint's `scrollableAncestor` prop](https://github.com/brigade/react-waypoint) and use it here, but we're probably better off just sticking to the defaults until someone requests it.

danoc:
Since I'm going on vacation soon and this needs some more QA, I'm going to close this and re-visit when I get back.

danoc:
@tomgenoni — This is ready for review. The `LazyImage` component is the same as the one used in `website` (`pro-list-result/components/lazy-image.jsx`).

PR #141, opened 2019-02-22
Clarify a few steps in RELEASING.md

danoc:
* Adds @webbower and @mateo42 to list of developers that can publish.
* Clarifies what to do about Lerna's private packages bug.
* Fixes an incorrect section heading


danoc:
Ah, you're right. I've changed that since it'll drive me nuts now that I know it's lowercase.

danoc:
While we're on the topic... https://github.com/thumbtack/thumbprint/pull/143

danoc:
@tomgenoni — Can you take another look?

PR #142, opened 2019-02-22
Add "centered grid" example

### LINKED ISSUE #28: [tp-ui-react-grid]: Add example for single column layout
Chatted with Dan about this offline. The React Grid has been simplified and doesn't support a common use case where a single Column of content should be centered in the grid. The recommendation is to use Atomic width and margin classes instead of a Grid.

tomgenoni:
Adds example of a "centered grid" using Atomic. 

Fixes #28 (assuming we're not building this into a component)

### DIFF
--- a/www/src/pages/components/grid/react/index.mdx
+++ b/www/src/pages/components/grid/react/index.mdx
@@ -68,6 +68,20 @@ import { ComponentHeader, ComponentFooter } from 'components/thumbprint-componen
 </Grid>
 ```
 
+### Centered grid
+
+Though this design pattern is often built using the Grid component, it is more easily solved by using Atomic classes: `mw7 center tc`.
+
+```html
+<div class="mw7 center tc">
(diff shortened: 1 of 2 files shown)


www/src/pages/components/grid/react/index.mdx
```diff
@@ -68,6 +68,20 @@ import { ComponentHeader, ComponentFooter } from 'components/thumbprint-componen
 </Grid>
 ```
 
+### Centered grid
+
+Though this design pattern is often built using the Grid component, it is more easily solved by using Atomic classes: `mw7 center tc`.
```
danoc:
```suggestion
The `Grid` component should not be used for centered single-column grids. This can be built with [Atomic classes](https://thumbprint.design/atomic/): `mw7 center tc`.
```
The "often built using the Grid component" isn't really true since the React component doesn't support it.

PR #144, opened 2019-02-22
Document all color tokens

danoc:
We previously marked some as "Private" so that they could only be used internally. @tomgenoni and I both think we should just document all of them for increased clarity.


danoc:
@tomgenoni — I made an issue to track removing the `private` property from the token JSON files (#145). Didn't want to include it in this PR because it'll trigger a tokens version bump I'd rather wait till I'm back to do that.

PR #154, opened 2019-03-05
Upgrade documentation packages to support IE 11

danoc:
This fixes https://github.com/thumbtack/thumbprint/issues/24.

Open related issues:
- [x] ~https://github.com/mdx-js/mdx/issues/433 (This is why this PR has changes to two MDX files.)~
- [x] ~https://github.com/ChristopherBiscardi/gatsby-mdx/issues/301 (There's a PR out to fix this issue. I've confirmed that it will fix our issue.)~


danoc:
This seems to be causing issues with the snapshots in Atomic since the version of `autoprefixer` got updated in one of our dependencies. Looking into it.

danoc:
Tried to fix the snapshot issue by generating the Atomic CSS like this:

```
#!/usr/bin/env node

const fse = require('fs-extra');
const sass = require('node-sass');
const postcss = require('postcss');
const autoprefixer = require('autoprefixer');
const cssnano = require('cssnano');
const mqpacker = require('css-mqpacker');
const nodeSassImporter = require('./node-sass-importer');

/**
 * Ensures that errors throw an error with a stacktrace.
 */
process.on('unhandledRejection', error => {
    throw error;
});

(async () => {
    const { css } = sass.renderSync({
        file: './src/atomic.scss',
        importer: nodeSassImporter,
    });

    const processedCss = await postcss([autoprefixer, cssnano, mqpacker]).process(css, {
        from: undefined,
    });

    await fse.outputFile('dist/atomic.css', processedCss);
})();
``

Thought that maybe the wrong version of Autoprefixer/PostCSS was getting used. Didn't seem to help.

danoc:
The snapshot changes are happening because the Gatsby update also causes a bump in the version of `browserslist` being used. 

Our config supports all of these browsers:
https://browserl.ist/?q=%3E+1%25+in+US%2Clast+2+versions%2Cie+11%2Cnot+dead

danoc:
@tomgenoni — Here's an example page that now works:
https://deploy-preview-154--thumbprint.netlify.com/components/modal-default/react/

PR #155, opened 2019-03-05
Truncation guidelines

tomgenoni:
This adds our philosophy around truncation, identifies the common patterns, and provides web-specific solutions where possible. Any native-specific solutions I will add in a follow up.


www/src/pages/guide/product/truncation/index.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+### 2. Multiline strings
+
+<div class="ba b-gray-300 pa4">
+    <img src="./multiline.png" alt="Multiline truncation example" width="300" />
+</div>
+
+#### Web
+
+This type of truncation is a bit more complicated with no one perfect solution. There a number of Sass and Javascript approaches but they buggy or convoluted. In this case we recommend a Sass approach that uses `line-height` tokens to calculate a `max-height`.
```
danoc:
```suggestion
This type of truncation is a bit more complicated with no one perfect solution. There are a number of Sass and JavaScript approaches but they are buggy or convoluted. In this case we recommend a Sass approach that uses `line-height` tokens to calculate a `max-height`.
```

Issue #38, opened 2019-01-26
Create `thumbprint-scss` and `thumbprint-global-css` package


danoc:
Following the recent consolidation of our React components into a single package, we're planning the same for our SCSS components. This has a few advantages:

- Simplifies dependencies in React and SCSS files
- Enables distribution of Thumbprint CSS files for non-React work
- Simplifies change, release, and update process—freeing up time to for Design Systems to build components.

Notes: 
- This work does not introduce any breaking changes. These are all code consolidations and swaps. No code will be removed.
- The `font-face` package will remain separate.
- The names/paths of packages may change slightly.

## New packages
We're creating the following 2 packages:

- [x] **`thumbprint-global-css`**:  Distributes:
1. `thumbprint-global.css`   Contains global styling for our reset, body, and default anchor styles. 

- [x] **`thumbprint-scss`**: Distributes:
1.  `thumbprint-mixin.scss`  Contains only mixins and our lone (deprecated) function. 
2.  `thumbprint-components.css`  Contains the output of all SCSS components — e.g., buttons, type, form elements. 

### Related tasks

- [x] Update documentation to work with the new file structure. (https://github.com/thumbtack/thumbprint/issues/37)
- [x] Update documentation to explain how to include CSS and SCSS dependencies.

danoc:
@tomgenoni — I think we can close this one.

tomgenoni:
I can move the migration task into another issue so the scope of this one isn't so big.

danoc:
That's probably best off for JIRA/some other tracker since it's website specific.

PR #160, opened 2019-03-07
Add docs for thumbprint-scss package

### LINKED ISSUE #91: Document `thumbprint-global-css` and `thumbprint-scss` packages
`thumbprint-scss` will render as they have before along side each React equivalent.

`thumbprint-css` is a new package. Could/should it be documented like Atomic?

tomgenoni:
This adds docs for `thumbprint-scss`

1. Added "Global SCSS" entry under "Global CSS"
2. Updated data query to filter out this new page
3. Ported text from "Mixins" and "Functions" entries and deleted them

Fixes #91, #38


tomgenoni:
@danoc is there a better name than "Global SCSS"?

danoc:
Maybe "SCSS Mixins" or "Mixins and Functions"?

It's a bit hard to read IMO since the two names are so similar:
image

Issue #156, opened 2019-03-06
Have modal focus go to the modal container, not an element within the modal


danoc:
I initially implemented `ModalCurtain` to focus on the first focusable element. While some articles say this is a best practice, it can be frustrating for users using screen-readers since the screen reader will skip over the modal contents.

danoc:
Here's a great article on modal accessibility:
https://developer.paciellogroup.com/blog/2018/06/the-current-state-of-modal-dialog-accessibility/

> When a modal dialog is activated, focus must be moved to the dialog. Where focus is initially placed may vary depending on the dialog’s content, but focusing the dialog itself can provide a consistently predictable user experience.

danoc:
Fixed in #163.

PR #163, opened 2019-03-08
Focus on `ModalCurtain` root instead of first focusable element (#156)

jichan-thumbtack:
- Upgraded Enzyme to easily test for `:focus` state (https://github.com/airbnb/enzyme/pull/1965)

- Upgraded forcus-trap-react (https://github.com/davidtheclark/focus-trap-react/blob/master/CHANGELOG.md) One breaking change seems to be that <FocusTrap> no longer wraps children in an extra div; instead it expects a single top-level child (better that it's one less nested `<div>`?)


packages/thumbprint-react/components/ModalCurtain/index.jsx
```diff
@@ -82,39 +84,53 @@ export default class ModalCurtain extends React.Component {
...
-                {shouldRenderEscListener && <EscListener onEscPress={onCloseClick} />}
+                {/**
+                    Use tabIndex="-1" to allow programmatic focus (as initialFocus node for <FocusTrap>)
+                    but not be tabbable by user.
+                */}
+                <div
+                    role="dialog"
+                    aria-label={accessibilityLabel}
+                    tabIndex="-1"
+                    ref={div => {
```
danoc: Would be great to use React's latest API for creating ref's:
https://reactjs.org/docs/refs-and-the-dom.html#creating-refs

jichan-thumbtack: Focus-trap-react only supports callback ref for now :( https://github.com/davidtheclark/focus-trap-react/blob/master/src/focus-trap-react.js#L82) Support for object ref is under CR (https://github.com/davidtheclark/focus-trap-react/pull/41) so we can switch after that goes through?

Another thing that I noticed when testing modals on screenreader was that we need aria-labelledby and aria-describedby for screenreader to announce contents in `ModalHeader` and `ModalDescription`. (https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html) Is this something already being worked on?

danoc: Thanks for digging into that.

After learning more about VoiceOver, I think this _is_ working. Try this:

1. Turn on VoiceOver and visit https://deploy-preview-163--thumbprint.netlify.com/components/modal-default/react/
2. Open a modal.
3. Hold Control + Option + Shift and then press the down arrow. This should take you to the title.
4. Hold Control + Option + Shift and then use the left and right arrow keys. This should cycle through the DOM nodes in the modal.

We could use `ModalHeader` and `ModalDescription` but there are plenty of times that people don't use those components and roll-their own styles.

danoc: One thing I noticed in my testing is that the modals are accessible to screen readers even when they're closed. I wonder if we should add `aria-hidden="true"` if it's closed.

That can be a separate task, but I'd be curious to know what you think. (I looked into `aria-modal` but was still able to navigate outside of the modal in VoiceOver 🤷‍♂️.)

PR #167, opened 2019-03-13
BlockList component

### LINKED ISSUE #73: Add React component of "block-list"
`tp-block-list` is being used ~12 times in `website`. We should consider making a React component.

cc: @bower

tomgenoni:
Fixes #73


packages/thumbprint-react/components/BlockList/index.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+BlockList.propTypes = {
+    /**
+     * Add a border to the bottom or surround the entire list.
+     */
+    border: PropTypes.oneOf(['bottom', 'group']),
+    /**
+     * Remove padding on left and right side of each list item.
+     */
+    flush: PropTypes.bool,
```
danoc: We've never used `tp-block-list--flush`. Should we leave it out until we have a use case?

packages/thumbprint-react/components/BlockList/index.module.scss
```diff
@@ -0,0 +1,35 @@
...
+.item {
+    padding: $tp-space__2 $tp-space__3;
+
+    &:not(:first-child) {
+        border-top: 1px solid $tp-color__gray-300;
+    }
+}
+
+.itemLink {
+    padding: $tp-space__2 $tp-space__3;
```
danoc: Should we bump up the specificity here to prevent people from overriding these values? Similar to this approach: https://github.com/thumbtack/thumbprint/blob/519ff414afd76aff9cb98d7bddf1487c62386f0f/packages/thumbprint-react/components/Type/index.module.scss#L4-L16

www/src/pages/components/block-list/react/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+## Default with complex link
+
+Note the use of `color-inherit` to override the default blue link color of `<BlockListItemLink>`.
+
+```jsx
+<BlockList border="group">
+    <BlockListItem>
+        <BlockListItemLink to="https://thumbtack.com" className="color-inherit hover-bg-gray-200">
+            <div class="flex">
```
danoc:
```suggestion
            <div className="flex">
```

www/src/pages/components/block-list/react/index.mdx
```diff
@@ -0,0 +1,127 @@
...
+                    We bring out the best in our colleagues — and our product — by pushing each
+                    beyond what we believe we’re capable of. We give honest feedback, even when it’s
+                    hard.
+                </Text>
+            </div>
+        </BlockListItemLink>
+    </BlockListItem>
+    <BlockListItem>
+        <BlockListItemLink to="https://thumbtack.com" className="color-inherit hover-bg-gray-200">
+            <div class="flex">
```
danoc:
```suggestion
            <div className="flex">
```

packages/thumbprint-react/components/BlockList/index.jsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import styles from './index.module.scss';
+
+export function BlockListItemLink({ children, to, className }) {
```
danoc: Instead of white-listing these three props, we could simply allow all props:

```jsx
export function BlockListItemLink({ className, ...rest }) {
    return (
        // eslint-disable-next-line jsx-a11y/anchor-has-content
        <a
            className={classNames({
                [styles.itemLink]: true,
                [className]: className !== undefined,
            })}
            {...rest}
        />
    );
}
```

The PropTypes can then become:

```js
BlockListItemLink.propTypes = {
    /**
     * URL pointing to the item link destination.
     */
    href: PropTypes.string.isRequired,
    /**
     * The content of the item link.
     */
    children: PropTypes.node.isRequired,
};
```

And examples would use `href` instead of `to`. I think this is a little better since this component is only concerned with the three lines of CSS that allow the developer to get a full clickable target.

packages/thumbprint-react/components/BlockList/index.jsx
tomgenoni: Sure.

packages/thumbprint-react/components/BlockList/index.module.scss
tomgenoni: I thought about that, couldn't decide. And there is at least one case where the padding was overridden. No strong feelings.

danoc: Probably worth doing so that we can more confidently refactor in the future. If it needs to be overwritten and there's a good use case, we could add that as a prop in the future.

packages/thumbprint-react/CHANGELOG.md
```diff
@@ -2,8 +2,7 @@
 
 ## Unreleased
 
-### Changed
```
danoc: The heading disappeared here

packages/thumbprint-react/components/BlockList/index.module.scss
tomgenoni: That would apply to every class in the file, yes?

packages/thumbprint-react/CHANGELOG.md
tomgenoni: Fixed

packages/thumbprint-react/components/BlockList/index.module.scss
danoc: It'd be specifically for `BlockListItemLink` since that's the only one where devs can pass in custom classes.

tomgenoni: Oh, of course. Thanks.

packages/thumbprint-react/components/BlockList/index.jsx
tomgenoni: Fixed

packages/thumbprint-react/components/BlockList/index.module.scss
danoc: The `!important` part is needed too so that Atomic can't override it.

packages/thumbprint-react/components/BlockList/test.jsx
```diff
@@ -0,0 +1,62 @@
...
+        expect(wrapper).toMatchSnapshot();
+    });
+    test('renders a link', () => {
+        const wrapper = mount(
+            <BlockListItemLink href="https://thumbtack.com">goose</BlockListItemLink>,
+        );
+        expect(wrapper.find('a').prop('href')).toBe('https://thumbtack.com');
+        expect(wrapper).toMatchSnapshot();
+    });
+    test('renders a link', () => {
```
danoc:
```suggestion
    test('forwards custom props', () => {
```

www/src/pages/components/block-list/scss/index.mdx
```diff
@@ -36,7 +38,7 @@ import { ComponentHeader, ComponentFooter } from 'components/thumbprint-componen
 <ul class="tp-block-list tp-block-list--all">
     <li>Item One</li>
     <li>Item Two</li>
-    <li><a href="#" class="tp-block-list__link">Item Fours</a></li>
+    <li>Item Three</li>
     <li>Item Three</li>
```
danoc:
```suggestion
     <li>Item Four</li>
```

www/src/pages/components/block-list/scss/index.mdx
```diff
@@ -47,7 +49,7 @@ import { ComponentHeader, ComponentFooter } from 'components/thumbprint-componen
 <ul class="tp-block-list-group">
     <li>Item One</li>
     <li>Item Two</li>
-    <li><a href="#" class="tp-block-list__link">Item Fours</a></li>
+    <li>Item Three</li>
     <li>Item Three</li>
```
danoc:
```suggestion
     <li>Item Four</li>
```

packages/thumbprint-react/components/BlockList/index.module.scss
tomgenoni: Added

PR #170, opened 2019-03-18
Use subset font

tomgenoni:
This change points the URL to our subsetted fonts which have a smaller character set, and are about 40% smaller, than the originals.

### DIFF
--- a/packages/thumbprint-font-face/CHANGELOG.md
+++ b/packages/thumbprint-font-face/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
--   [Minor] Change font path to point to subsetted version of font.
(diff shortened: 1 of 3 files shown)


packages/thumbprint-font-face/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
+-   [Minor] Change font path to point to subsetted version of font.
```
danoc:
```suggestion
-   [Major] Change font path to point to subsetted version of font.
```

Would technically be a breaking change since it requires the font to exist on `$thumbprint-font-url`. Won't actually require an upgrade for internal Thumbtack users if we put the font on our server.

PR #175, opened 2019-03-21
Alert color change for "warning" variation

tomgenoni:
Changes background color to a faint yellow and the text color dark.


danoc:
Thanks for doing this! Looks infinitely better.

PR #179, opened 2019-03-26
Add legal notice to brand assets page

danoc:
Adding per request from our team.


danoc:
Added WIP because we might tweak it a little.

danoc:
@tomgenoni — This is ready for review.

PR #182, opened 2019-03-26
Allow `isSticky` to work at all breakpoints

### LINKED ISSUE #135: Make `ModalFooter` `isSticky` prop work at all breakpoints
Currently the `isSticky` prop only affects the modal when rendered at small breakpoints. After a discussion with @jonpkerwin and @dcapo, we want to change it so that the sticky footer applies to all breakpoints.

The `ModalDefault` component would remain 100% width and height on mobile devices. When `isSticky` is `true` though, the modal height will now be 100% except for the top and bottom padding.

danoc:
This fixes #135.

To review this, I'd recommend first reading the `CHANGELOG.md` file then reading the `*.jsx` files.

FWIW, I'm not crazy about the prop and `className`'s that I used in this PR but I couldn't think of anything better. Open to suggestions/feedback.

I think this is a backwards compatible change, but I'll be QA'ing our modals before shipping this in website.

You can preview this here:
https://deploy-preview-182--thumbprint.netlify.com/components/modal-default/react/

***

cc: @mxpan


danoc:
Found a regression that I need to fix. The "Tall modal without a sticky footer" lost its bottom border under the close button. This is what it should look like:

image

danoc:
Just fixed that issue. ⬆️

danoc:
@dcapo — Just fixed the modal height issue you mentioned on Slack.

danoc:
@dcapo — Had to fix a merge conflict. Could you take another look?

Preview is live:
https://deploy-preview-182--thumbprint.netlify.com/components/modal-default/react/

danoc:
@mxpan — When do you need this by? @dcapo — LMK if there's anything I can do to help you test this.

mxpan:
> @mxpan — When do you need this by?

Ideally by end of next week!

PR #183, opened 2019-03-27
Create codemods package.

danoc:
This is a port from code in a separate repo. By putting it here and publishing to NPM, we can make it easier to use because consumers will be able to do `npx @thumbtack/thumbprint-codemods` and use the CLI without needing to clone a repo.

It also allows us to benefit from the lint, testing, and pretter infra that this codebase already has.


danoc:
I'm doing this now because I'm going to make a codemod for the `Button` secondary to tertiary migration.

PR #186, opened 2019-03-27
Remove tp spacer dependency

### LINKED ISSUE #159: Remove all usage of `tp-spacer` from our codebase
We've deprecated it a while ago but we still need to remove it from internal Thumbprint code.

tomgenoni:
Fixes #159 

I checked the components visually but would be good if you took a look at Modal and any other high-importance components.


tomgenoni:
I removed the `mixins` from the files that no longer needed it.

PR #187, opened 2019-03-27
Create a React Image component

danoc:
### Todos
- [x] Explore how media queries would work
- [x] How will placeholders work?
- [x] Support `png` and `gif`.
- [x] Should we have a `isFullWidth` prop that prevents it from calculating sizes?
- [x] How can `ResponsiveImage` be combined with `LazyImage`?
- [x] What does it look like for a developer that wants to use an image from our image service?
- [x] How will `object: fit` work? Is it a prop?
- [x] Should `AspectRatio` be its own component?
- [x] Add image service sizes as Tokens?
- [x] Change to not watch image width
- [x] Remove gif support
- [x] Should we spread props into the `img`
- [x] How will this work with `Avatar` and `ServiceCardImage`
- [x] Add CHANGELOG entry
- [ ] Add tests
- [ ] Add more examples

### Consider
- [x] optional `width` prop for fixed-width images
- [x] possible to remove `<span style="font-size: 0px;"></span>`?
- [x] `className` for easier positioning, margin, etc


packages/thumbprint-react/components/Image/components/aspect-ratio/index.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+const AspectRatio = ({ children, ratio }) => (
```
tomgenoni: Any reason not to add the aspect ratio calc to the `<picture>`? Would make the HTML a little cleaner. https://tomgenoni.github.io/experiments/aspect-ratio/

packages/thumbprint-react/components/Image/components/aspect-ratio/index.jsx
```diff
@@ -0,0 +1,25 @@
...
+    <div
+        style={{
+            height: 0,
+            position: 'relative',
+            paddingBottom: `${ratio * 100}%`,
+        }}
+    >
+        {children({
+            childStyles: {
+                position: 'absolute',
```
tomgenoni: I think you can get away with only `position: absolute` and `top: 0`...?

packages/thumbprint-react/components/Image/components/aspect-ratio/index.jsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+const AspectRatio = ({ children, ratio }) => (
+    <div
+        style={{
+            height: 0,
```
tomgenoni: You shouldn't need `height:0`.

www/src/pages/components/image/react/index.mdx
```diff
@@ -0,0 +1,109 @@
...
+import { ComponentHeader, ComponentFooter } from 'components/thumbprint-components';
+
+<ComponentHeader data={props.data} />
+
+## Examples
+
+### Magic Image
+
+```jsx
+<MagicImage id="350357374752325652" aspectRatio={864 / 1600} />
```
tomgenoni: Not sure how to handle this but there are two scenarios for `aspectRatio`:

1. The consumer wants an image in one of our available [aspect ratios](https://thumbprint.design/guide/product/aspect-ratio/). 
2. The consumer wants an image, isn't requesting a specific aspect ratio, but wants to hold the space.

They are different concepts though not sure if it's worth trying to decouple them. I can imagine a consumer requesting an unofficial aspect ratio, like 5/6, and wondering why the image service is not returning an image in that aspect ratio (it would only return the original aspect ratio).

packages/thumbprint-react/components/Image/components/image/index.jsx
```diff
@@ -0,0 +1,248 @@
...
+        <>
+            <Waypoint
+                onEnter={() => {
+                    setShouldLoadImage(true);
+                    onEnter(true);
+                }}
+                // Start loading slightly before the user scrolls to the image
+                topOffset={-100}
+                bottomOffset={-100}
+            >
```
tomgenoni: Could we add `horizontal` prop here? Passed through with `horizontalLazyload` from `SmartImage`. That would unlock carousels. Tested locally by hardcoding and it worked as expected.

danoc: I'm looking into removing React Waypoint in favor of `react-intersection-observer` with a polyfill. That would remove the need for the prop. It seems that `horizontal` exists because of performance concerns: https://github.com/civiccc/react-waypoint/issues/66#issuecomment-198870080

(If we were to add the prop, we'd also have to add it to `Avatar` and `ServiceCardImage` which is 🤮.)

danoc:
Tested on BrowserStack and got these results:

**Issues:**
* Safari on iPhone 6S with iOS 9: Wrong PNG size loaded (got `1024px` since it was in the `img` `src`) and [object-position didn't work](https://caniuse.com/#feat=object-fit).
* Safari 9.1 on OS X El Capitan: `object-position` doesn't work
* Internet Explorer 11 on Windows 8.1: Loaded PNG in the wrong size but `object-position` on it did work well. Issue may be related to the object-fit polyfill.

**Working well:**
* Google Nexus 6
* Samsung Galaxy Note 4 (2014)
* Motorola Moto G 2nd Gen (2014)
* Internet Explorer 11 on Windows 7
*  Safari on iPhone 7 with iOS 10
*  Safari on iPhone SE with iOS 11
*  Chrome on Google Nexus 5 (2013)
* Firefox 48 on OS X Snow Leopard
* Safari 11.1 on macOS High Sierra

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -0,0 +1,234 @@
...
+     * Crops the image at the provided height. The `objectFit` and `objectPosition` props can be
+     * used to control how the image is cropped.
+     */
+    height: PropTypes.string,
+    /**
+     * Crops the image to a specific aspect ratio and creates a [placeholder box](https://css-tricks.com/aspect-ratio-boxes/)
+     * for the image. The placeholder prevents the browser scroll from jumping when the image is
+     * lazy-loaded.
+     */
+    aspectRatio: PropTypes.string,
```
mateo42: IMHO, this prop should be a decimal representation of the ratio rather than a string.  While there's a readability aspect to a string, it's also more limiting and introduces a format concern that isn't really relevant to this component.  Any consumer should be capable of providing the decimal version, and in some situations might end up having to convert from another format to colon separated, and back.

danoc: I initially had it as a decimal, but changed due to feedback from @tomgenoni. Agree that the decimal is simpler, but I can see it being confusing since `ImageServiceProvider` uses strings. I'm not sure which approach I prefer.

mateo42: I would argue that ImageServiceProvider is a separate concern from this component and os it shouldn't be enforced that Image uses the same format.   I think it makes sense for ImageServiceProvider to take a string since it supports a limited set of ratios, but the either 
1) the aspect ratio is passed to render props as a decimal or

danoc: Yeah, I think that's a fair point. @tomgenoni — what do you think?

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Converts `"8:5"` to `8/5`.
+ */
+const aspectRatioStringToFloat = aspectRatio => {
+    if (!aspectRatio) {
+        return null;
+    }
+
+    const arr = aspectRatio.split(':');
+    return parseFloat(arr[0]) / parseFloat(arr[1]);
```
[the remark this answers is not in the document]
danoc: Just added something, but will wait until the string versus decimal discussion is done before adding tests for it.

packages/thumbprint-react/components/Image/index.jsx
tomgenoni: I understand the point but my concern was that we have same named props in close proximity that require different values for reasons that are complicated to explain (and no one will remember). My preference is that we don't confuse an already complicated set of components in this way.

Could it be a fraction `8/5`? The decimal requires determining the width/height of the image, doing math and then how many decimal places should I use...? Seems like extra work.

mateo42: > My preference is that we don't confuse an already complicated set of components in this way.

Completely agree. do you think renaming the passthrough prop on ImageServiceProvider to `placeholderRatio` (a decimal value) would solve this?  Ideally you would not need to remap any props for most use cases:

```
<ImageServiceProvider id="123456" aspectRatio="16:9">
    // imageProps is an object that includes { placeholderRatio: 1.7777 }
    {(imageProps) => <Image {...imageProps} />}
</ImageServiceProvider>
```

mateo42: In this case `<ImageServiceProvider>` does the logic of converting from 16:9 to 1.7777. `<Image>` will only check `0 < placeholderRatio < 2`.

danoc:
Just renamed to `containerAspectRatio` and made it take a number, not a string.

PR #206, opened 2019-04-04
Remove tp-spacer

tomgenoni:
This removes `functions.scss` which contained our lone `tp-spacer` function. Also removed references to it in the docs.


packages/thumbprint-scss/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Removed
+
+-   [Major] Remove `function.scss` that contained the `tp-spacer` function.
```
danoc: Could you add a note here on how to upgrade? Something along the lines of use `$tp-space-*` Tokens and hardcode values that don't fit in that system.

danoc:
cc: @zboman, @benknight

packages/thumbprint-scss/CHANGELOG.md
tomgenoni: Added.

tomgenoni:
Goodnight, sweet prince

benknight:
Looks like `tp-spacer` is not used on the help repo.

PR #207, opened 2019-04-04
Replace tp-spacer-unit usage

tomgenoni:
Replaces the couples instances `$tp-spacer__unit` token with a hardcoded pixel value.


tomgenoni:
@danoc had a merge conflict. can you re-review?

PR #217, opened 2019-04-17
Add isFullBleed prop to ModalDefault

### LINKED ISSUE #194: Consider supporting a full bleed ModalDefault
Currently, ModalDefault has strict ownership of its content's padding. This prevents children from having a full bleed background color, as seen here:

Storybook

Maybe we can consider supporting some sort of padding override to allow for this use case?
Designer: johan

benknight:
I took an open approach here by creating a new prop at the parent ModalDefault component level. Setting this boolean prop to true cancels padding entirely on in the content container and leaves spacing up to the consumer of ModalDefault.  I believe this is the best approach as it makes no assumptions about the use cases for allowing content to bleed to the edge, making it possible for child content to extend to the left, right, and bottom edge of the modal. 

I didn't add any new examples because this is kind of an escape hatch and thus I don't think it constitutes a primary use case for ModalDefault, and it's easy enough to test the behavior this prop by adding it to the existing examples in the live code.

Closes #194

### DIFF
--- a/packages/thumbprint-react/CHANGELOG.md
+++ b/packages/thumbprint-react/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+-   [Minor] Add `isFullBleed` prop to the `ModalDefault` component. (#194)
+
 ## 0.5.3 - 2019-04-09
 
 ### Added
--- a/packages/thumbprint-react/components/ModalDefault/__snapshots__/test.jsx.snap
+++ b/packages/thumbprint-react/components/ModalDefault/__snapshots__/test.jsx.snap
@@ -2,6 +2,7 @@
 
 exports[`ModalDefault does not show the close button if \`shouldHideCloseButton\` is true 1`] = `
 <ModalDefault
+  isFullBleed={false}
   isOpen={false}
   onCloseClick={[Function]}
   shouldCloseOnCurtainClick={true}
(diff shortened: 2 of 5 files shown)


danoc:
Hi Ben! Thanks for the PR.

The main case we want to support is allowing developers to extend a background-color and border to take up 100% of the modal width (#194). I think this approach makes that difficult for a few reasons:

* It breaks the footer padding when the footer is not sticky.
* The developer would have to re-add padding to the `ModalHeader`. This can lead to visual inconsistencies because the padding wouldn't get updated if we change the `ModalDefault` padding.
* We have a `ModalDefaultAnimatedWrapper` component that is similar to this prop. The use cases overlap a bit. (We don't document that component because we generally want to discourage it.)

Have you considered an in-between approach where the user can reliably mark specific parts of the content as full-bleed?

Here's an example:

image

An example of an implementation would be to create a component (`ModalDefaultContentFullBleed`) that does the padding and negative margins. It would accept custom styles and classes allowing the developer to add background colors and borders and stuff.

benknight:
The proposed solution in the PR was based on considering the use case mentioned in #194 in addition to my own specific use case of showing an iframe inside ModalDefault that acts as a portal into the Help Center, extending it all the way to the left, right, and bottom edge of the padding.

Now that I see `ModalDefaultAnimatedWrapper` looks to be more appropriate for that use case, I agree that the correct solution is something tied to the specific use case in #194 where specific content sections can extend out to the left and right edges.

PR #223, opened 2019-04-27
Add `onMonthChange`, `daysThemeDotIndicator`, and `daysThemeStrikeout` props to `DatePicker`

komatsu:
https://thumbtack.atlassian.net/browse/PX-73


packages/thumbprint-react/components/DatePicker/index.jsx
```diff
@@ -118,4 +122,16 @@ DatePicker.propTypes = {
...
+     * buttons or via keyboard. The function receives a single Javascript Date object indicating
+     * the new on-screen month.
+     */
+    onMonthChange: PropTypes.func,
+    /**
+     * Allows for custom styles and interaction with day cells in the calendar. See
+     * https://github.com/gpbl/react-day-picker/blob/v6.2.1/docs/docs/modifiers.md for supported
+     * inputs and behavior.
+     */
+    modifiers: PropTypes.shape({}),
```
danoc: What's the desired product goal with this prop? Do we need all of the functionality that modifiers supports? We could possibly bake the functionality into the component while keeping the API small. A few possible benefits:

* Keeping the API small makes it easier to move away from react-day-picker in the future. 
* When modifying the component in the future, we'll have greater confidence that our changes won't break the component. It's not clear to me, for example how `disabledDays` interacts with `modifiers`.
* The `DatePicker` component is a component that is optimized for Thumbtack's needs. If we need a different styling for disabled days, for example, then it should be baked into the component and not left for consumers to style themselves.

komatsu: The final styles are still up-in-the-air based on what we can support in this component. Ideally, we want to be able to support several scenarios:
1. Crossed-out (but still selectable) days
2. Some indicator dot to indicate when we have relevant events on the calendar for any given date. Initial designs had 1, 2, or 3 dot indicators depending on the number of events, but we may change this to just a single dot.
3. Grayed-out (but still selectable) days, e.g. weekends/non-business hour days. We may not use this yet, but it was in the initial set of designs.
4. bolding of "today"

The ability to pass either a function or an object describing the set of dates seems useful but that complicates things a bit. I also think that the term 'disabledDays' becomes a bit ambiguous given these other states that we would want. 

Should we start by adding something along the lines of:
- strikeoutDays
- indicatorDays
- isTodayBolded

komatsu: Per discussion with @danoc , we'll be adding two new props in place of exposing the entire modifier:
- daysThemeDotIndicator -- accepts function. Adds a dot indicator below the date if true.
- daysThemeStrikeout --accepts function. Strikes out date if true.

// Modify `modifiers` prop that gets passed into react-day-picker.
// modifiers['theme-dot'] = theDotFunction
// modifiers['theme-strikeout'] = theStrikeoutFunction

// Add css styles for --dot and --strikeout. 
.DayPickerDay--theme-dot
.DayPickerDay--theme-strikeout

packages/thumbprint-react/components/DatePicker/index.jsx
```diff
@@ -65,8 +65,23 @@ export default class DatePicker extends React.Component {
...
+            lastMonth,
+            onMonthChange,
+            daysThemeDotIndicator,
+            daysThemeStrikeout,
+        } = this.props;
         const { selectedDays } = this.state;
+        const modifiers = {};
+
+        if (typeof daysThemeDotIndicator === 'function') {
+            Object.assign(modifiers, { 'theme-dot': daysThemeDotIndicator });
```
danoc: Can we do this instead since `Object.assign` doesn't work in IE 11?

```js
modifiers['theme-dot'] = daysThemeDotIndicator;
```

PR #224, opened 2019-04-29
Upgrade documentation packages

danoc:
Routine package upgrade. Includes a breaking change to MDX now that MDX is on v1.

**Edit:** This also bumps up snapshots because `browserlist`/`autoprefixer`/`caniuse-lite` got updated since they are dependencies of dependencies.

### DIFF
--- a/packages/thumbprint-scss/__snapshots__/test.js.snap
+++ b/packages/thumbprint-scss/__snapshots__/test.js.snap
@@ -2,7 +2,6 @@
 
 exports[`compiles correctly 1`] = `
 ".tp-alert {
-  display: -webkit-flex;
   display: flex;
   font-size: 14px;
   line-height: 20px;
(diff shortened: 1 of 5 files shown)


packages/thumbprint-scss/__snapshots__/test.js.snap
```diff
@@ -2,7 +2,6 @@
 
 exports[`compiles correctly 1`] = `
 ".tp-alert {
-  display: -webkit-flex;
```
tomgenoni: Is this an issue?

danoc: Brought this up on Slack in `#front-end-infra` to see who has concerns.

I confirmed that not having flex prefixes is the correct behavior for our `browserslist` config. We happened to still be adding them because our packages were old so the browser usage data was stale.

danoc: Got a 👍to kill the prefixes.

PR #234, opened 2019-05-08
Unmount ModalDefault after the tests run

danoc:
The modals don't automatically get cleaned up after each test, so it was causing code from the early tests to leak into the later tests.


danoc:
@tomgenoni — Could you take a look at this one when you get a chance? Tests all still pass 👍

PR #236, opened 2019-05-09
Remove "name" field from tokens

danoc:
The field isn't used anywhere and will become especially duplicative when I follow up with a PR that removes the category name from the id.

**To-do:**

- [x] Add changelog entry
- [x] Fix failing documentation build


danoc:
@tomgenoni — I made a bit of a mess with the PRs, but this one is now ready for review.

PR #240, opened 2019-05-16
Remove references to deprecated tokens.

danoc:
Snapshots didn't change which means that the SCSS ones are a 1:1 mapping.

### DIFF
--- a/packages/thumbprint-scss/mixins/type.scss
+++ b/packages/thumbprint-scss/mixins/type.scss
@@ -90,72 +90,72 @@
 @mixin tp-heading-1($important: null) {
     font-size: 32px $important;
     line-height: 1.3 $important;
-    font-weight: 500 $important;
+    font-weight: $tp-font__weight__normal $important;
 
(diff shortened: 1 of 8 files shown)


packages/thumbprint-scss/mixins/type.scss
```diff
@@ -88,74 +88,74 @@
 // DEPRECATED
 
 @mixin tp-heading-1($important: null) {
-    font-size: $tp-font__heading__1__size $important;
-    line-height: $tp-font__heading__1__line-height $important;
-    font-weight: $tp-font__heading__1__weight $important;
+    font-size: 32px $important;
+    line-height: 1.3 $important;
+    font-weight: 500 $important;
```
tomgenoni: `500` can be `400` since our Mark font doesn't have that weight and is loading that now anyway.

danoc: Oh, right. Forgot about that. Will fix.

PR #249, opened 2019-05-21
Update dot centering on DatePicker

senseiurata:
The current implementation uses a hacky
media query to compensate for the shrinking
grid items. This update implements true
centering.


senseiurata:
Current:
https://cl.ly/b9a5d005ee41

After:
https://cl.ly/e4c548f4fd03

PR #250, opened 2019-05-21
Update date picker design

senseiurata:
The CSS / design is from another existing location in Thumbtack. It will introduce a breaking change, because the day item will not shrink beyond 44px. At 320px, there is only 12px left because 308px = 44px * 7 days. The calendar will fit on the screen but we'll need to update existing usages to let it bleed over the typical padding value (16px) we have on the outer layout (which seems like a reasonable trade-off at small viewport width). 

The 44px touch target is a great improvement from the current 31px-39px and seems like a good number to aim for, based on Target Size article on WCAG 2.1:
https://www.w3.org/WAI/WCAG21/Understanding/target-size.html


packages/thumbprint-react/components/DatePicker/index.module.scss
```diff
@@ -47,11 +35,11 @@
...
 .root :global(.DayPicker-NavButton) {
-    top: 130px;
-    border: 1px solid $tp-color__gray;
+    top: 0;
+    border: 0;
     position: absolute;
-    width: 40px;
-    height: 90px;
+    width: 75px;
+    height: 1.5rem;
```
danoc: What's the benefit of using `rem` for the `height` but pixels for the `width`? Would the arrow get stretched vertically?

danoc:
This is looking good! I've pinged @jonpkerwin on Slack for design feedback.

packages/thumbprint-react/components/DatePicker/index.module.scss
senseiurata: I made some more improvements:
- I converted few things into px because the text size is in px. The vertical alignment with the month button breaks when the units are inconsistent (the button stretches but the text doesn't).
- Increased the touch target of the month button to 44px.

danoc:
The code looks good to me! @jonpkerwin — Would you be able to do a visual design pass on this today?

This is the preview:
https://deploy-preview-250--thumbprint.netlify.com/components/date-picker/react/

And this is the current one:
https://thumbprint.design/components/date-picker/react/

jonpkerwin:
Did a visual pass and looks good to me! Thanks!

Jon

senseiurata:
The PR is now updated with the requested design. Since this was a larger change than simply borrowing styles from IR pro list calendar, I took the opportunity to take a deeper dive (as indicated in my commit message):
> - Convert non-px value to px for consistency
> - Delete .DayPicker-WeekNumber (unused for Thumbprint)
> - Move order of .DayPicker-Day--outside so the border
>   does not appear if a day outside the calendar month
>   is selected
> - Remove cursor: pointer on buttons. See
>   https://www.w3.org/TR/css-ui-3/#valdef-cursor-pointer
> - 48px x 48px touch targets, with 44px min-width

The last two points are proposals, and I'm particularly thinking that the cursor change might be the most controversial. I think that this article makes a strong case against putting a cursor on non-links:

https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b

tl:dr; it's intentional that pointers are not used on buttons, and [Apple](https://developer.apple.com/library/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/Pointers.html) / [Microsoft](https://docs.microsoft.com/en-us/windows/desktop/uxguide/inter-mouse) guidelines, as well as [W3C spec](https://www.w3.org/TR/css-ui-3/#valdef-cursor-pointer) all suggest that pointers should specifically be used on links.

danoc:
Nice work, as always! 👏 

I'm open to discussing the button cursor topic, but I think that should be separate form this PR. If we did make that change, we'd want to make it for multiple components, not just `DatePicker`.

danoc:
I'll sync with @jonpkerwin and Cory for design QA.

packages/thumbprint-react/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
+-   [Patch] Update `DatePicker` component design.
```
```suggestion
-   [Major] Update `DatePicker` component design.
```

senseiurata: Updated!

PR #252, opened 2019-05-23
Add image fade to onload

### LINKED ISSUE #243: Add quick fadeIn prop to Image component
When `onLoad` is triggered this would add a fast opacity fade in, around 300ms. 

This can be on by default but consumers should have the ability to disable it as there may be use cases where this undesired.

tomgenoni:
Adds a 300ms fade in on the image so the loading isn't so abrupt. 

Will publish as a beta for testing.

Fixes #243

### DIFF
--- a/packages/thumbprint-react/components/Image/components/picture.jsx
+++ b/packages/thumbprint-react/components/Image/components/picture.jsx
@@ -1,14 +1,29 @@
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useState } from 'react';
+import styles from './picture.module.scss';
 
 const Picture = forwardRef((props, ref) => {
     const { src, sources, sizes, alt, style, ...rest } = props;
+
+    const [loadedClass, setClass] = useState(null);
+
+    function handleOnLoad() {
+        setClass(styles.imageLoaded);
+    }
 
     return (
         <picture>
             {sources.map(({ type, srcSet }) => (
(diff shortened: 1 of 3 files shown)


packages/thumbprint-react/components/Image/components/picture.jsx
```diff
@@ -1,14 +1,29 @@
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useState } from 'react';
+import styles from './picture.module.scss';
 
 const Picture = forwardRef((props, ref) => {
     const { src, sources, sizes, alt, style, ...rest } = props;
 
+    const [loadedClass, setClass] = useState(null);
```
danoc: This is a bit easier to read IMO if it's:

```js
const [isLoaded, setIsLoaded] = useState(false);
```

And then in the `img`:

```jsx
<img
    src={src}
    alt={alt}
    style={style}
    ref={ref}
    {...rest}
    onLoad={() => { setIsLoaded(true) }}
    className={isLoaded ? styles.imageLoaded : styles.imageLoading}
/>
```

I think it works better since `isLoaded` more accurately describes the "state" of the image whereas `loadedClass` is very specific to what we intend to do with that state.

tomgenoni: Yeah, that's cleaner. Just updated.

tomgenoni:
@danoc this PR seems to be in a weird review state. Are there unresolved changes you've requested?

PR #254, opened 2019-05-23
Pill component

tomgenoni:


danoc:
@tomgenoni — is this ready for review? Feel free to make it and add me as a review. Looks good to me. I also made sure that this could work with `Tooltip` and it works fine if you have `Tooltip > TextButton > Pill`.

PR #263, opened 2019-06-03
Create an iOS output for tokens

danoc:
This adds iOS tokens for `Color` and `Space`. It also sets up CocoaPods for publishing the new tokens.

## Notes on the PR

* The `packages/thumbprint-tokens/src/tokens/*.json` files are the source of truth.
* The JSON files get converted to Swift with [Handlebars](https://handlebarsjs.com/). 
  1. `src/index.js` is the entrypoint
  2. `src/templates/ios.handlebars` is the template
  3. `src/helpers/ios.js` has functions that the template uses.
* The `src/__snapshots__/index.test.js.snap` file stores the output of the generated file. This is helpful when reviewing future changes to Tokens infrastructure.

## To-do

- [x] Confirm that output is the same as the code in the iOS codebase.
- [x] Write documentation for publishing Tokens for iOS.
- [x] Add comments to Podspec.


packages/thumbprint-tokens/src/__snapshots__/index.test.js.snap
```diff
@@ -362,3 +362,105 @@ $tp-wrap__no-pad-width: 1010px;
...
 `;
+
+exports[`outputted iOS matches snapshot 1`] = `
+"import UIKit.UIColor
+
+/// These are the colors we use at Thumbtack.
+public enum Color {
+    /// Maps to #eaf6fa
+    public static let blue100: UIColor = UIColor(red: 0.91764706, green: 0.9647059, blue: 0.98039216, alpha: 1.0)
+    /// Maps to #b3ebff
```
kevinmbeaulieu: nit: Can we add an empty line between each property (e.g., in between lines 372 & 373 here)? Makes it easier to read when inspecting the source during iOS . development

packages/thumbprint-tokens/src/__snapshots__/index.test.js.snap
```diff
@@ -362,3 +362,105 @@ $tp-wrap__no-pad-width: 1010px;
 $tp-z-index__modal: 200;
 "
 `;
+
+exports[`outputted iOS matches snapshot 1`] = `
+"import UIKit.UIColor
+
+/// These are the colors we use at Thumbtack.
+public enum Color {
+    /// Maps to #eaf6fa
```
kevinmbeaulieu: nit: Can we preserve the comment structure currently used in the iOS repo? (e.g., `Blue 100 – #eaf6fa.`) Just a little more formal

packages/thumbprint-tokens/src/__snapshots__/index.test.js.snap
```diff
@@ -362,3 +362,105 @@ $tp-wrap__no-pad-width: 1010px;
...
+    /// Maps to #e9eced
+    public static let gray300: UIColor = UIColor(red: 0.9137255, green: 0.9254902, blue: 0.92941177, alpha: 1.0)
+    /// Maps to #d3d4d5
+    public static let gray: UIColor = UIColor(red: 0.827451, green: 0.83137256, blue: 0.8352941, alpha: 1.0)
+    /// Maps to #ffffff
+    public static let white: UIColor = UIColor(red: 1.0, green: 1.0, blue: 1.0, alpha: 1.0)
+}
+
+public enum Space {
+    public static let one: CGFloat = 4.0
```
kevinmbeaulieu: Can we add documentation comments to these as well? (e.g., `Spacer 1 – 4.0pt.` is what we're currently using in the iOS repo) Having the raw value in the documentation helps because it gets shown during autocomplete so you don't have to memorize all the name => raw value mappings when coding)

packages/thumbprint-tokens/src/index.js
```diff
@@ -2,12 +2,34 @@
...
+const JSZip = require('jszip');
 const handlebars = require('handlebars');
 
 const outputs = [
     { slug: 'javascript-cjs', distName: 'index.js' },
     { slug: 'javascript-es', distName: 'index.js' },
     { slug: 'scss', distName: '_index.scss' },
+    {
+        slug: 'ios',
+        distName: 'index.swift',
```
kevinmbeaulieu: Any objections to renaming this to `ThumbprintTokens.swift`? More standard in Swift naming conventions than `index`

packages/thumbprint-tokens/src/tokens/color.json
```diff
@@ -4,212 +4,350 @@
...
         {
             "id": "blue-300",
-            "value": { "web": "#79d2f2" },
-            "type": "color"
+            "value": {
+                "web": "#79d2f2",
+                "ios": "UIColor = UIColor(red: 0.4745098, green: 0.8235294, blue: 0.9490196, alpha: 1.0)"
+            },
+            "type": "color",
+            "private": true
```
kevinmbeaulieu: What's the `"private": true` property on some of the colors?

danoc: Whoops. Added that by accident. It's something we used to have in our tokens set-up for colors that we wanted to reserve for use only within Thumbprint. We've since gotten rid of that concept.

danoc:
@kevinmbeaulieu — This is ready for another look.

danoc:
Thanks for the feedback, @kevinmbeaulieu!

@tomgenoni — Can you take a look at this?

PR #264, opened 2019-06-03
Add `onClick` and `shouldOpenInNewTab` props to ServiceCard

zboman:
As per my [discussion on Slack](https://thumbtack.slack.com/archives/C7FLM0ZGU/p1559588809025200) with @tomgenoni, here are the changes to support new `onClick` and `target` props on `ServiceCard`. I welcome any and all feedback. 👍


tomgenoni:
@zboman Mind adding an example of these props to `www/src/pages/components/service-card/react/index.mdx?

packages/thumbprint-react/components/ServiceCard/index.jsx
```diff
@@ -50,9 +50,9 @@ function ServiceCardDescription({ iconColor, icon, children }) {
     );
 }
 
-export default function ServiceCard({ url, children }) {
+export default function ServiceCard({ url, onClick, target, children }) {
```
danoc: Thoughts on calling the `target` prop `shouldOpenInNewTab`? A boolean may be better since we want to support `_blank` and the default `_self`, not `_parent` ot `_top`. (This is also consistent with the prop name in `Button`.)

zboman: Good call. I found the `getAnchorProps` utility that was being used in `Plain` and decided that it would be a good idea to employ here as well so I did and pushed that up. Hope that was okay. The only side effect is that it also provides an `isDisabled` prop, but the behavior will be consistent between `ServiceCard` and `Plain`/`Link`, which I think is a net positive. Thoughts?

zboman:
@tomgenoni I added an example. Let me know if it needs tweaking.

tomgenoni:
@zboman Thanks for adding the example. Looks good. Only concern is the undocumented `isDisabled` prop but sounds like you and @danoc are discussing this.

zboman:
@tomgenoni @danoc I think this is ready for final review. I updated the PR with a small change to make icons align better when the `ServiceDescription` contains more than one line of text. And I tried to document the `isDisabled` prop by listing it in the `propTypes` and by providing it in the example code. But if we want to document it better, I'm all open to suggestions. Or I can remove it if needed. 👍

danoc:
Hm, I think we'd be better off either:

* Implementing the `shouldOpenInNewTab` manually without relying on the code in `UIAction`.
* Moving the shared code (minus the `isDisabled` part) into [`/utils`](https://github.com/thumbtack/thumbprint/tree/master/packages/thumbprint-react/utils) and updating both `UIAction` and `ServiceCard` to point to it.

I think there are two advantages to those approaches:

* By not including `isDisabled`, we keep `ServiceCard`'s API as small as possible. (We can always add it later if we find a good use case for it!)
* When importing across folders in `/components`, we generally only import "index" files (`index.js`/`index.jsx`) and consider the other files to be part of the "private API" of a component. This separation makes the codebase easier to maintain because an it means that an engineer working in `/components/UIAction` can make refactors of the `UIAction` private API without worrying about breaking components that don't use the `Themed` or `Plain` components that it exports.

I don't have a strong opinion on which of those two options is best though.

zboman:
Thanks @tomgenoni @danoc

PR #265, opened 2019-06-03
Add giles to contributors

lavelle:
Just going through the flow to make sure I have git and everything else set up


danoc:
Woo! Welcome back. 🎉

PR #267, opened 2019-06-04
Add dynamic colours to initial avatars

lavelle:


packages/thumbprint-react/components/Avatar/index.jsx
```diff
@@ -33,6 +34,29 @@ const getBadgeProps = ({ size, hasUnreadNotifications, isChecked, isOnline }) =>
...
+    tokens.tpColorYellow100,
+    tokens.tpColorRed100,
+    tokens.tpColorPurple100,
+    tokens.tpColorBlue100,
+];
+
+const getBackgroundColour = initials =>
+    BACKGROUND_COLOURS[initials.charCodeAt(0) % BACKGROUND_COLOURS.length];
+
+const getTextColour = initials => TEXT_COLOURS[initials.charCodeAt(0) % TEXT_COLOURS.length];
```
danoc: `initials` is not a required prop, so this will fail if it's not defined. I think that's why Jest is failing. Let's use `gray200` as a default background color in the off chance that neither an initial or an image are provided.

If you'd like, we can also use `warning` to add a warning for that case.

packages/thumbprint-react/components/Avatar/index.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import * as tokens from '@thumbtack/thumbprint-tokens';
```
danoc: Does Rollup handle this type of import well? You can check by running `yarn run build:dist` in the Thumbprint root folder and then checking the contents of `packages/thumbprint-react/dist` to make sure that it didn't pull in all of the tokens as part of the bundling.

packages/thumbprint-react/components/Avatar/index.jsx
```diff
@@ -33,6 +34,29 @@ const getBadgeProps = ({ size, hasUnreadNotifications, isChecked, isOnline }) =>
...
+    tokens.tpColorPurple100,
+    tokens.tpColorBlue100,
+];
+
+const getBackgroundColour = initials =>
+    BACKGROUND_COLOURS[initials.charCodeAt(0) % BACKGROUND_COLOURS.length];
+
+const getTextColour = initials => TEXT_COLOURS[initials.charCodeAt(0) % TEXT_COLOURS.length];
+
 class EntityAvatar extends React.Component {
```
danoc: This change should apply to both the `Avatar` and `EntityAvatar` components.

packages/thumbprint-react/components/Avatar/index.jsx
lavelle: Done

packages/thumbprint-react/components/Avatar/index.jsx
lavelle: it already does

packages/thumbprint-react/components/Avatar/index.jsx
lavelle: seems like rollup is smart. the dist directly it gets compiled to this

```
import { tpColorGray200, tpColorWhite, tpColorIndigo600, tpColorGreen600, tpColorYellow600, tpColorRed600, tpColorPurple600, tpColorBlue600, tpColorIndigo100, tpColorGreen100, tpColorYellow100, tpColorRed100, tpColorPurple100, tpColorBlue100 } from '@thumbtack/thumbprint-tokens';
```

lavelle:
Think i've addressed everything

packages/thumbprint-react/components/Avatar/index.jsx
danoc: Ah, wow. That's neat. 👍

packages/thumbprint-react/components/Avatar/index.jsx
danoc: Ah, I misread the diff.

packages/thumbprint-react/components/Avatar/index.jsx
```diff
@@ -33,6 +34,32 @@ const getBadgeProps = ({ size, hasUnreadNotifications, isChecked, isOnline }) =>
...
+const BACKGROUND_COLOURS = [
+    tokens.tpColorIndigo600,
+    tokens.tpColorGreen600,
+    tokens.tpColorYellow600,
+    tokens.tpColorRed600,
+    tokens.tpColorPurple600,
+    tokens.tpColorBlue600,
+];
+
+const TEXT_COLOURS = [
```
danoc: You got the text and background colors mixed up. Should be dark text on a light background.

***

Also, the `BACKGROUND_COLOURS` and `TEXT_COLOURS` would be easier to maintain if they were combined into a single `COLOURS` variable that is an array of objects like `{ text: '#fff', background: '#000' }`. This makes the relationship between the background and text color more explicit.

lavelle: Ah yeah, my bad. Combined the arrays and flipped the fg/bg

lavelle:
Oops sorry, keep forgetting about those pesky snapshots

PR #269, opened 2019-06-04
Add modal default content full bleed option

### LINKED ISSUE #194: Consider supporting a full bleed ModalDefault

lavelle:
Fixes #194.


packages/thumbprint-react/components/ModalDefault/index.jsx
```diff
@@ -160,6 +160,21 @@ const modalContentPropTypes = {
...
 
+const modalContentFullBleedPropTypes = {
+    /**
+     * Content (usually a form) that makes up the main part of the modal.
+     */
+    children: PropTypes.node.isRequired,
+    /**
+     * Allows the React `className` prop to be passed through to the rendered element.
+     */
+    className: PropTypes.string.isRequired,
```
danoc: `className` and `style` shouldn't be required because you can use one but not the other.

packages/thumbprint-react/components/ModalDefault/index.module.scss
```diff
@@ -204,6 +204,13 @@ $width-wide: 1400px;
     padding-top: 12px;
 }
 
+// The double selectors in this file ensure that consumers using the `className` prop can't easily
+// overwrite the styles that we want to enforce in the component.
+.modalContentFullBleed.modalContentFullBleed {
+    margin: 0 -32px !important;
```
danoc: This is causing horizontal scrolling on mobile. 
horizontalscroll

Haven't thought through the possible solutions much, but `overflow-x: hidden` may work.

packages/thumbprint-react/components/ModalDefault/index.module.scss
```diff
@@ -204,6 +204,13 @@ $width-wide: 1400px;
...
+    padding: 0 32px !important;
```
danoc: Let's use a variable that is shared with `ModalContents` so that this is more likely to stay in sync in the future.

packages/thumbprint-react/components/ModalDefault/index.jsx
lavelle: done

packages/thumbprint-react/components/ModalDefault/index.module.scss
lavelle: done

packages/thumbprint-react/components/ModalDefault/index.module.scss
lavelle: good catch. the padding actually needed to be responsive just like the padding for the main content element. still learning all the edge cases for this component.

lavelle:
Should be ready for another look

danoc:
@lavelle — Just realized that this approach won't work for the screenshot in #194. The buttons in the screenshot scroll horizontally, so that part needs to be 100% width (without the padding) so that they get cut off at the edge of the screen, not within the box afforded to `children`.

Maybe we do need two components?

PR #270, opened 2019-06-05
Add verticalAlign prop to Checkbox and Radio (#266)

### LINKED ISSUE #266: [@thumbtack/thumbprint-react]: Custom alignment for <Checkbox> and <Radio>
In `<Checkbox>` and `<Radio>` components, the root containing input and text (label) are always vertically center-aligned.
(i.e. https://github.com/thumbtack/thumbprint/blob/master/packages/thumbprint-react/components/Checkbox/index.module.scss#L6)

I've seen several designs where we would want them to be aligned at the start, especially in case where text wraps over multiple lines. Previously, I got around this by having two separate `<Checkbox>` component and vanilla `<label>` tag but seeing that this is a recurring pattern, what do you think about adding `align` prop to`<Checkbox>`/`<Radio> `components whose value may be either 'center' or 'start' (open to naming suggestions). I can put up a PR if this is okay. 

screenshot-www figma com-2019 06 03-17-50-16
screenshot-www figma com-2019 06 03-18-00-09

jichan-thumbtack:
* Adds `checkboxVerticalAlign` prop to `<Checkbox>`
* Adds `radioVerticalAlign` prop to `<Radio>`

Fixes #266

### DIFF
--- a/packages/thumbprint-react/CHANGELOG.md
+++ b/packages/thumbprint-react/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+-   [Minor] Add `checkboxVerticalAlign` prop to `Checkbox` and `radioVerticalAlign` prop to `Radio` (#266)
+
 ## 2.1.0 - 2019-06-04
 
 ### Added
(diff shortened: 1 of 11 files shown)


danoc:
I'll merge and release this before lunch!

tomgenoni:
@jichan-thumbtack Thank you for this. Much appreciated.

PR #271, opened 2019-06-05
Rename Avatar to UserAvatar

lavelle:


lavelle:
This implements the rename – tests and docs still look good. Now I just need to look into the codemods stuff.

lavelle:
Sure, dunno how much in flight stuff there is but that sounds fine. done.

PR #273, opened 2019-06-05
Add a `button-secondary-to-tertiary` codemod

danoc:
This does a few things:

* Removes the old `thumbprint-react-consolidation` codemod that didn't use jscodeshift. It's not needed anymore and I'd rather have us use the same tech as the other open-source codemods.
* Creates a `button-secondary-to-tertiary` codemod. (@bgando found some bugs in my version. This should be updated to use her code.)
* Makes a light CLI that will let us run codemods like this: `npx @thumbtack/thumbprint-codemods **/*.jsx --codemod 'button-secondary-to-tertiary'`. The alternative [is this process](https://github.com/pinterest/gestalt#codemods).

## To-do:
- [ ] Write a README.md showing how to write a codemod and how to run tests locally.


danoc:
@bgando - would you be able to replace the buggy button codemod here with your version?

danoc:
@bgando — Can you please take a look? 🙇‍♂️

PR #274, opened 2019-06-05
Add delay after hovering before showing tooltip

### LINKED ISSUE #158: Add a open delay on the tooltip
It currently opens on hover which is jarring for users that accidentally hover on the tooltip.

Here's a great article about this:
https://www.nngroup.com/articles/timing-exposing-content/

This is their recommendation:

> For hover interactions, the timing guidelines for speed of visual feedback and exposing hidden elements must be broken down into more steps:
>
> 1. Mouse cursor enters target area: display visual feedback within 0.1 seconds.
> 2. Wait 0.3–0.5 seconds.
> 3. If cursor remains stopped within target area, display corresponding hidden content within 0.1 seconds.
> 4. Keep displaying the exposed content element until the cursor has left the triggering target area or the exposed content for longer than 0.5 seconds.

lavelle:
Fixes #158

### DIFF
--- a/packages/thumbprint-react/CHANGELOG.md
+++ b/packages/thumbprint-react/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Changed
 
+-   [Patch] Add small delay before showing tooltip after hovering to prevent flickering. (#158)
 -   [Patch] Redesign the avatar with initials design to use dynamic colours. (#25)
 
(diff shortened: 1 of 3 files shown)


lavelle:
This feels good to me. You can move your cursor over quickly without it appearing, but if you are trying to open it it still feels snappy.

PR #277, opened 2019-06-05
Add codemod and tests for the Avatar component rename

lavelle:


packages/thumbprint-codemods/src/avatar-import-name/test.js
```diff
@@ -0,0 +1,73 @@
...
+            transform(`
+import React from 'react';
+import { Avatar, Button, Input } from '@thumbtack/thumbprint-react';
+const App = (props) => (
+    <div>
+        <Avatar />
+        <Button />
+        <Input />
+    </div>
+ )`),
```
danoc:
```suggestion
)`),
```

lavelle: done

lavelle:
> This is great. Their utility function makes it so easy.
> 
> Does this work well if:
> 
> * `@thumbtack/thumbprint-react` is not one of the imports?
> * `@thumbtack/thumbprint-react` is an import but `Avatar` is not imported?

Could you give an example of these tests? Surely if we're not importing `Avatar` then the code should be a no-op, right?

danoc:
@lavelle — Just pushed a failing test case.

This code:

```jsx
import React from 'react';
import Avatar from '../my-avatar.jsx';
const App = (props) => (
    <Avatar />
)`;
```

...is transformed to:
import React from 'react';
import UserAvatar from '../my-avatar.jsx';
const App = (props) => (
    <UserAvatar />
)`;
```jsx

```

PR #279, opened 2019-06-06
Use `Image` component instead of lazysizes in `ServiceCardImage`.

danoc:


packages/thumbprint-react/components/ServiceCard/index.jsx
```diff
@@ -1,32 +1,18 @@
...
-        return (
-            <div className={`${styles.image} lazyload`} data-bg={url} role="img" aria-label={alt} />
-        );
-    }
-}
+const ServiceCardImage = ({ url, sources, alt }) => (
+    <Image
+        className={styles.image}
+        sources={sources}
+        containerAspectRatio={8 / 5}
```
tomgenoni: Should we enforce this? Don't have strong feelings but if a user uses a different aspect ratio they may not notice the placeholder is incorrect. E.g, they request "16:9" but that "8/5" is what's in the placeholder it'll still "work" but just the placeholder will be wrong.

danoc: Hm, I say we should keep it since it makes the lazy-loading placeholder work. We'd otherwise have to add a prop to `ServiceCardImage` that is the aspect ratio of the image which seems like overkill since we decided when designing the component that we only want to support `8:5` images.

tomgenoni: Will the user have to add `aspectRatio="8:5"` to `ImageServicePicture`? If that's the case my question was why not let `containerAspectRatio` get auto-populated. 

The concern was that the user might add `aspectRatio="16:9"` (for an experiment) which, unless they were paying close attention, would appear to work but in fact the placeholder size is incorrect. So unless I'm missing something it's not really enforcing an aspect ratio.

danoc: The functionality of the aspect ratio in this PR is the same as it was in the existing version of `ServiceCardImage`. Both versions use CSS to enforce an aspect ratio. If a user were to pass in a `1:1` image, `ServiceCardImage` would still render it as `8:5` with `object-fit: cover`.

Issue #89, opened 2019-02-01
Change Button's `secondary` theme, add `tertiary` theme


danoc:
Our buttons are getting a small makeover. The `secondary` theme (gray text) will become `tertiary`. There is a new `secondary` theme that has blue text.

***
Here are the new styles:

![image](https://user-images.githubusercontent.com/916038/52094806-6789db80-2575-11e9-8243-5c2ed5640142.png)

***

### Migration plan

1. ~Create the `tertiary` theme that is exactly the same as our current `secondary` style. Use [`warning`](https://www.npmjs.com/package/warning) package to show that `secondary` is deprecated. Release as a non-breaking change.~ https://github.com/thumbtack/thumbprint/pull/93
2. Write script to replace `<Button theme="secondary" />` and `<ThemedLink theme="secondary" />` with  `<Button theme="tertiary" />` and `<ThemedLink theme="tertiary" />`. (Started work in the [`daniel/codemod`](https://github.com/thumbtack/thumbprint/compare/daniel/button-codemod) branch.)
3. Perform the migration.
4. Create Thumbprint PR to change secondary theme from the old gray to the new blue. Release as breaking change.
5. In websites using Thumbprint, ensure no more `theme="secondary"` exists. Upgrade the `@thumbtack/thumbprint-react` package.

PR #281, opened 2019-06-06
Add new `secondary` theme for Button and ThemedLink components

bgando:
New secondary theme:
image

#89


danoc:
Also I made this issue to track removing the alerts in 2 weeks or so: #282.

PR #283, opened 2019-06-06
Improve README and simplify maintence of codemods

danoc:
The `index.js` list of codemods are now dynamic. No need to update a hard-coded list.


bgando:
LGTM

PR #286, opened 2019-06-07
Create codemod for Avatar size prop breaking changes

lavelle:


lavelle:
Ready for review. PTAL.

packages/thumbprint-codemods/src/avatar-sizes/test.js
```diff
@@ -0,0 +1,167 @@
...
+        import { Avatar } from './my-avatar.jsx';
+        const App = (props) => (
+            <>
+                <Avatar size="large" />
+                <Avatar size="medium" />
+                <Avatar size="small" />
+            </>
+        )`;
+
+        expect(transform(code)).toBeNull();
```
danoc: Should this be `null`?

We do it will `null` in the `Button` test but not in the other `Avatar` test:
https://github.com/thumbtack/thumbprint/blob/master/packages/thumbprint-codemods/src/avatar-import-name/test.js#L75-L95

I'm not sure which one is correct.

packages/thumbprint-codemods/src/avatar-sizes/test.js
```diff
@@ -0,0 +1,167 @@
...
+                <Avatar size="small" />
+                <Input />
+            </div>
+        )`;
+
+        expect(transform(codeBefore)).toBe(codeAfter);
+        expect(console.error).toHaveBeenCalledTimes(1);
+    });
+
+    test.only('props are spread and also has a large size', () => {
```
danoc: Forgot the `.only` here. 😄

packages/thumbprint-codemods/src/avatar-sizes/test.js
```diff
@@ -0,0 +1,167 @@
...
+                <>
+                    <A size="medium" />
+                    <A size="small" />
+                    <A size="xsmall" />
+                </>
+            )`);
+    });
+});
+
+describe('output should change but also throw a console error', () => {
```
danoc: Another test case that I think would be good to add:

```js
test('throws error if props are spread after a size is defined', () => {
    const codeBefore = `import React from 'react';
    import { Avatar, Input } from '@thumbtack/thumbprint-react';
    const App = (props) => (
        <div>
            <Avatar size="large" {...props} />
            <Input />
        </div>
    )`;

    const codeAfter = `import React from 'react';
    import { Avatar, Input } from '@thumbtack/thumbprint-react';
    const App = (props) => (
        <div>
            <Avatar size="medium" {...props} />
            <Input />
        </div>
    )`;

    expect(transform(codeBefore)).toBe(codeAfter);
    expect(console.error).toHaveBeenCalledTimes(1);
});
```

We care if the props are spread after the size, but don't care if props are spread and the `size` is defined after the spread.

packages/thumbprint-codemods/src/avatar-sizes/test.js
lavelle: done

packages/thumbprint-codemods/src/avatar-sizes/test.js
lavelle: yeah, the behaviour is different. the other avatar codemod returns the source code even when it doesn't change anything, whereas this one and the button one return null. should probably update the avatar name codemod to have the same behaviour as here, but that's a different PR

packages/thumbprint-codemods/src/avatar-sizes/test.js
lavelle: good test. added

lavelle:
ready for another look

packages/thumbprint-codemods/src/avatar-sizes/test.js
danoc: Cool, just wanted to see which was one correct.

danoc:
Should this one include `EntityAvatar` or do you want to do that as something separate?

lavelle:
I would rather do as a follow-up, to keep the scope of the commit down. It would also require more tests

PR #295, opened 2019-06-10
Enable lodash linting. Fixes #293

### LINKED ISSUE #293: Enable eslint-plugin-lodash to catch some browser API bugs at build time
We enabled `eslint-plugin-compat` to catch uses of browser methods that are not available in old but supported browsers like IE11. However, this has some false negatives – it does not catch prototype methods like `Array.find` and `String.startsWith`.

`eslint-plugin-lodash` can catch these, although it can also flag some false positives for other methods with those names – eg. a custom class `X` with its own `find` method would be flagged due to the limitations of static analysis of untyped code.

I propose we enable this plugin with a conservative configuration:

- All the "error" lint rules – these are uncontroversial
- None of the "stylistic" ones – these don't catch browser bugs, and can cause arguments about preferred style
- Enforce the import syntax to only allow `import method from 'lodash/method';` type imports, and ban `import _ from 'lodash'` which will give you the monolithic build and massively bloat our bundles.
- Enable the `prefer-lodash-method` rule, but *only* for methods that are not supported in IE11. Others, like `Array.map` which are available everywhere can still be used. Here is my list of methods I know to be unsupported in IE11. Please update the issue if you find more:
   - ` _.find`
    - `_.includes`
    - `_.startsWith`
    - `_.values`
    - `_.entries`

All of these rules would only be enabled for the `thumbprint-react` package – any other code in the repo that is not distributed to browsers can use whatever it likes since it only runs in Node.

cc @danoc let me know if you agree with all that.

lavelle:


packages/thumbprint-react/components/DatePicker/index.jsx
```diff
@@ -46,7 +47,7 @@ export default class DatePicker extends React.Component {
 
         if (allowMultiSelection) {
             if (selected) {
-                const selectedIndex = newSelectedDays.findIndex(selectedDay =>
+                const selectedIndex = findIndex(newSelectedDays, selectedDay =>
```
danoc: Whoa, I wonder if this just never worked in IE 11? 🥴

This PR needs a CHANGELOG entry for Thumbprint React since it changes the source code.

lavelle: Yep, that would throw an error since that method is not available. The power of linting!

lavelle:
Updated changelog

PR #299, opened 2019-06-10
Refactor out common logic of codemods

lavelle:


.eslintrc.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
         'prettier',
         'prettier/react',
     ],
-    plugins: ['jest', 'react-hooks'],
+    plugins: ['jest', 'react-hooks', 'compat'],
```
danoc: Seems like this got some unrelated commits?

PR #300, opened 2019-06-10
Make textarea text black when there is an error but component is focused. Fixes #298

### LINKED ISSUE #298: Textarea text color should be black when focused and only red when blurred
Currently it is always red. Behaviour should match Input which is correct. There should never be red text and a blue border.

lavelle:


lavelle:
PTAL

PR #301, opened 2019-06-11
Image refactor to use img as placeholder

tomgenoni:
- Removes the placeholder `div`, applies same `padding-top` technique to the `` instead
- Removes the outer wrapper `div` making the `<picture>` the outer most element
- Removes reliance on absolute positioning on the image 
- Fixes bug where Edge 18 was downloading both `jpeg` and `webp` with responsive images. This may be a side effect of way the code is now organized.


packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -1,21 +1,33 @@
...
 import { useInView } from 'react-intersection-observer';
-import Picture from './components/picture.jsx';
 import canUseDOM from '../../utils/can-use-dom';
+import styles from './index.module.scss';
+
+// --------------------------------------------------------------------------------------------
+// Steps in rendering Image
+//
+// 1. Picture is rendered without src, srcSets, and with a padding-top placholder on the <img>
+// based on the aspectRatio or containerAspectRatio.
```
```suggestion
// based on the containerAspectRatio.
```

Let's just list `containerAspectRatio` since `aspectRatio` isn't a prop in this component?

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -24,30 +36,15 @@ const Image = forwardRef((props, outerRef) => {
...
-            // passing a `border-radius` in `className` or `style` would not work since the
-            // container is a `div`, not an `img`.
-            overflow: 'hidden',
-            // Setting a `min-height` makes the lazy-loading work in cases where the `Image` parent
-            // has a `0px` height.
-            minHeight: '1px',
-        },
-    };
+    // Used by srcSet to determine which image in the list will be requested. Unless the image
+    // is full width this value has to be calculated client-side because we don't know the
```
danoc: > Unless the image is full width

We don't currently provide a way to opt-out of this calculation, so it's always client-side.

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -139,55 +137,97 @@ const Image = forwardRef((props, outerRef) => {
...
-            right: 0,
-            bottom: 0,
-            left: 0,
-            height: '100%',
+            height: 0,
         };
     }
 
+    // --------------------------------------------------------------------------------------------
+    // Sources and scrSets
```
```suggestion
    // Sources and srcSets
```

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -139,55 +137,97 @@ const Image = forwardRef((props, outerRef) => {
...
-            {/* This `div` holds the space set by `containerAspectRatio`. */}
-            {Object.keys(aspectRatioBoxProps).length > 0 && <div {...aspectRatioBoxProps} />}
-
-            {shouldLoadImage && <Picture {...pictureProps} />}
+        <>
+            <picture
+                {...rest}
+                className={classNames(styles.picture, className)}
+                ref={el => {
+                    // Using a callback `ref` on this `div` allows us to have multiple `ref`s on one
```
```suggestion
                    // Using a callback `ref` on this `picture` allows us to have multiple `ref`s on one
```

packages/thumbprint-react/components/Image/index.jsx
tomgenoni: Ok, removed.

danoc:
https://deploy-preview-301--thumbprint.netlify.com/components/service-card/react/

The `ServiceCardImage` examples seem to be getting cut off. Haven't dug into why though ^

tomgenoni:
The `ServiceCardImage` is actually not getting cut off but is now showing the full image. The issue is we're sending it an image that's not in a 8:5 aspect ratio and because the image is no longer using `absolute` to fill the space created by the parent the full image is showing. There's no way to force the image into that aspect ratio without using `height` (which we can't because it changes). 

I think our options are:

1. Migrate `ServiceCard` images as the same time we release `Image` in `website` to use `ImageServicePicture` so we can send it an image in the correct aspect ratio. Maybe we whitelist `8:5` in `ServiceCardImage` to only accept that size.
2. Revert `ServiceCardImage` to not use `Image` and instead render the image as a background-image (which is how it was built originally to handle these odd image sizes) until we can update it properly via step 1.

tomgenoni:
Believe this is ready for review. My last couple commits... 

- Reverted to code prior to merging your branch
- Removed `containerAspectRatio` `object-fit` example
- `object-fit` now only added if `height` prop is populated
- Updated docs on what `containerAspectRatio` does now
- Updated `ServiceCardImage` to use `8-5`
- Updated changelog with breaking changes

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -84,7 +82,9 @@ const Image = forwardRef((props, outerRef) => {
     // Object Fit: polyfill and CSS styles
     // --------------------------------------------------------------------------------------------
 
-    const shouldObjectFit = height || containerAspectRatio;
+    const objectFitProps = {};
+
+    const shouldObjectFit = height;
```
```suggestion
    const shouldObjectFit = !!height;
```

This was in the old code too, but I just noticed now.

packages/thumbprint-react/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+### Changed
+
+-   [Patch] Simplify `Image` component by remove placeholder `div` and CSS positing so the component behaves more predictably. Fixes double download bug in Edge 18.
+-   [Major] Remove `object-fit` CSS from `Image` that was previously being added when the `containerAspectRatio` was present. This prop now only adds placeholder spacing on the `<img>` which is removed onload.
```
```suggestion
-   [Major] Only include `object-fit` CSS in `Image` when `height` is provided. That CSS was previously being added when the `containerAspectRatio` was present. `containerAspectRatio` now only adds placeholder spacing on the `<img>` which is removed `onload`.
```

Changed it to be a bit more clear that object-fit CSS sometimes is still included.

packages/thumbprint-react/CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
+-   [Major] Remove CSS that enforced aspect ratio in `ServiceCardImage` due to changes in `Image` component that no longer support this use case.
+
+## 5.0.0 - 2019-06-13
```
danoc: There's some other stuff here that isn't relevant to this PR. Everything below this should be unchanged.

packages/thumbprint-react/components/Avatar/index.jsx
```diff
@@ -148,8 +148,7 @@ EntityAvatar.defaultProps = {
     size: 'small',
 };
 
-// TODO(giles): remove this default export once website has been updated to refer only to UserAvatar
-export default class UserAvatar extends React.Component {
+class UserAvatar extends React.Component {
```
danoc: Unrelated to this PR.

packages/thumbprint-react/components/Image/index.jsx
danoc: Fix was removed in the reverts.

packages/thumbprint-react/components/Image/index.jsx
danoc: Fix was removed in the reverts.

packages/thumbprint-react/components/Image/index.jsx
danoc: Fix was removed in the reverts.

packages/thumbprint-react/CHANGELOG.md
tomgenoni: Updated with `master`.

packages/thumbprint-react/components/Image/index.jsx
tomgenoni: Fixed

packages/thumbprint-react/components/Image/index.jsx
danoc: Fix was removed in the reverts.

packages/thumbprint-react/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Changed
+
+-   [Patch] Simplify `Image` component by remove placeholder `div` and CSS positing so the component behaves more predictably. Fixes double download bug in Edge 18.
+-   [Major] Only include `object-fit` CSS in `Image` when `height` is provided. That CSS was previously being added when the `containerAspectRatio` was present. `containerAspectRatio` now only adds placeholder spacing on the `<img>` which is removed `onload`.
+-   [Major] Remove CSS that enforced aspect ratio in `ServiceCardImage` due to changes in `Image` component that no longer support this use case.
```
danoc: Need to add docs to `ServiceCardImage` that says the image must be `8:5`.

packages/thumbprint-react/components/Image/index.jsx
tomgenoni: Fixed

packages/thumbprint-react/components/Avatar/index.jsx
tomgenoni: Fixed

packages/thumbprint-react/CHANGELOG.md
tomgenoni: Added

tomgenoni:
Yes, Monday works.

PR #302, opened 2019-06-11
Newsletter for June 11

tomgenoni:


danoc:
@tomgenoni — I just added a bunch of stuff. I'm realizing that starting with the name of the person makes it hard to scan the list of tasks.

What if it was:

```
-   **<Subject>**: <Details> (<Contributor>)
```

Or:

```
-   <Details> (<Contributor>)
```

What do you think? Feel free to change it to whatever you think is more legible.

tomgenoni:
Yeah, I think it's good we give contributors credit (esp outside contributors) but I agree the emphasis should be on the task itself. Thanks for doing that.

danoc:
Sweet, that reads a lot better!

PR #314, opened 2019-06-13
Remove Avatar alias for UserAvatar

lavelle:


lavelle:
should be good to go. we can release 5.0.0 and bump in website later today

PR #319, opened 2019-06-15
Deprecate ServiceCardImage url in favor of src

tomgenoni:
Do have a way of rendering a `deprecated` tag for props? Looks like we can via https://github.com/thumbtack/thumbprint/blob/e9016e6b8589710641dbd7ded6434095dab5cdf0/www/src/components/thumbprint-components/component-footer/index.jsx#L41 but wasn't sure what the doclet syntax looked like, couldn't find an existing example.


danoc:
Just fixed the docs issue. I think it broke in a recent `react-docgen` breaking change. 🤷‍♂

PR #335, opened 2019-06-18
Fix border color of disabled `Button` with "caution" theme. (Fixes #325)

### LINKED ISSUE #325: Disabled caution button in SCSS has wrong border color
It looks like this:

image

... but it should look like this:

image

lavelle:


lavelle:
PTAL!

PR #339, opened 2019-06-20
Create popover component

lavelle:
This creates an initial version of the popover component, to attempt to lock down the API and get the visuals pixel-perfect. Once we have merged this, I will create another PR to make it interactive. This will be done by refactoring out a shared base component from the existing tooltip, which can then be wrapped by both Popover and Tooltip.

Known issues:

- [x] All the imports are not appearing in the docs
- [x] Not all nubbin positions are currently available in the API. Wanted to get feedback on the API before adding repetitive code for all 12 of them


lavelle:
This should be ready for review. I also added the usage page

lavelle:
Notes on things we may want to add for our meeting tomorrow:

- ESC to close
- Click outside to close
- Test accessibility / keyboard navigation
- Overflows / positioning after window resize
- Animation to fade in/out
- Portals/displacing
- Disable for SSR

lavelle:
Another thing to consider: passing a ref directly, rather than a function prop that renders the launcher

See https://github.com/thumbtack/thumbprint/pull/341/files

lavelle:
More known issues:

- [ ] Nubbin does not rotate after popper updates the position of the popover relative to the target

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+import { Themed } from '../UIAction/index.jsx';
+import { NavigationCloseTiny } from '../../icons/index.jsx';
+
+import styles from './index.module.scss';
+
+const ESC_KEY = 27;
+
+// Internal component only. Proptypes are defined for the main component `Popover` at the end of the
+// file.
+// eslint-disable-next-line react/prop-types
```
danoc: Let's also add `PropTypes` for the internal one. (I don't mind if we disable for tests, but I do think it's good to have for internal components.)

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+     */
+    children: PropTypes.string.isRequired,
+    /**
+     * Function called when the button is clicked.
+     */
+    onClick: PropTypes.func.isRequired,
+};
+
+const PopoverSecondaryButton = ({ children, onClick }) => (
+    <div className={styles.popoverButton}>
```
danoc: Seems like this class doesn't exist. Do we need the `div`?

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+
+PopoverBody.propTypes = {
+    /**
+     * The body description text for the popover.
+     */
+    children: PropTypes.string.isRequired,
+};
+
+const PopoverPrimaryButton = ({ children, onClick }) => (
+    <div className={styles.popoverButton}>
```
danoc: Seems like this class doesn't exist. Do we need the `div`?

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+    children: PropTypes.string.isRequired,
+};
+
+const PopoverBody = ({ children }) => <div className={styles.popoverBody}>{children}</div>;
+
+PopoverBody.propTypes = {
+    /**
+     * The body description text for the popover.
+     */
+    children: PropTypes.string.isRequired,
```
danoc: It's possible that someone would include a link within the text. Should it be a `node` instead?

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+    };
+
+    useEffect(() => {
+        document.addEventListener('keyup', handleKeyUp);
+        return () => {
+            document.removeEventListener('keyup', handleKeyUp);
+        };
+    });
+
+    Popover.handleClickOutside = () => onCloseClick();
```
danoc: Can this be simplified to:

```js
Popover.handleClickOutside = onCloseClick; 
```

packages/thumbprint-react/components/Popover/index.module.scss
```diff
@@ -0,0 +1,77 @@
+@import '~@thumbtack/thumbprint-tokens/dist/scss/_index';
+@import '~@thumbtack/thumbprint-scss/mixins';
+
+.root {
+    background: $tp-color__white;
+    box-shadow: 0px 2px 10px rgba(0, 0, 0, 0.2);
+    padding: $tp-space__3 $tp-space__4;
```
danoc: Can you check with @jonpkerwin on what he wants for the padding? The values here are different than they are in Figma.

packages/thumbprint-react/components/UIAction/themed.module.scss
```diff
@@ -196,6 +196,33 @@ $button-border-width: 2px;
...
+            border-color: $tp-color__indigo-500;
+        }
+    }
+
+    &ThemePopoverSecondary {
+        color: $tp-color__black-300;
+        background-color: $tp-color__white;
+        border-color: $tp-color__white;
+        padding: 0 10px;
+        font-weight: 400;
```
danoc:
```suggestion
        font-weight: $tp-font-weight__normal;
```

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+
+PopoverSecondaryButton.propTypes = {
+    /**
+     * The text for this button.
+     */
+    children: PropTypes.string.isRequired,
+    /**
+     * Function called when the button is clicked.
+     */
+    onClick: PropTypes.func.isRequired,
```
danoc: This may sometimes be a link to another page. (Image a "Learn more" CTA.)

Could use a `to` prop like: https://thumbprint.design/components/link/react/#section-themed-links

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+                })}
+                style={style}
+                data-placement={placement}
+            >
+                {children}
+
+                <div className={styles.closeButton}>
+                    <TextButton
+                        accessibilityLabel="Close popover"
+                        iconLeft={<NavigationCloseTiny className={styles.closeButtonIcon} />}
```
danoc: The click target is too small to be accessible. You can work around this by adding padding in `styles.closeButtonIcon`. For an example, see how `ModalDefault` does it.

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+                    : popoverContent)}
+        </Manager>
+    );
+};
+
+Popover.propTypes = {
+    /**
+     * Contents for the Popover. Usually a `PopoverTitle`, `PopoverBody`, and `PopoverPrimaryButton`
+     */
+    children: PropTypes.node,
```
danoc: Let's make this required?

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,219 @@
...
+        'left',
+        'left-end',
+        'right-start',
+        'right',
+        'right-end',
+    ]),
+    /**
+     * Whether or not the popover is visible.
+     */
+    isOpen: PropTypes.bool.isRequired,
```
danoc: Most (all?) of our controlled components make this not required and then default to `false` (or `undefined` for text inputs).

packages/thumbprint-react/components/Popover/index.jsx
lavelle: i agree you shouldn't have undocumented props, but in this case they're identical to the ones below, which is why i left a comment. anyone who wants to know what to pass in here can easily read those proptypes. IMO duplicating all of them is worse, it just adds visual noise to the file

packages/thumbprint-react/components/Popover/index.module.scss
lavelle: I'll ping him. The ones in figma were not all standard tokens units so I used the closest options.

packages/thumbprint-react/components/Popover/index.jsx
lavelle: Done

lavelle:
Addressed all the feedback, and QA'ed with Jon. Ready for another look!

packages/thumbprint-react/components/Popover/index.jsx
danoc: This one still seems to be an issue.

packages/thumbprint-react/components/Popover/index.jsx
```diff
@@ -0,0 +1,223 @@
...
+    // Appends the tooltip right before `</body>` when true.
+    const shouldDisplace = container === 'body';
+
+    const handleKeyUp = event => {
+        if (event.keyCode === ESC_KEY) {
+            onCloseClick();
+        }
+    };
+
+    useEffect(() => {
```
danoc: We only care about running this on mount and unmount, right? If so, we should add `[]` as the second argument to `useEffect`.

jichan-thumbtack:
This looks great! API makes senses and seems easy to use. 

A couple of questions I had:
- Does it support dismissal on clicking outside the popover? Or was this intentionally not supported?
- Is there anyway that Popover has configurable width? I feel like there maybe a lot of use cases where `max-width` of 300px is too limiting. 
- nit, but was wondering why the `<PrimaryButton?>` doesn't support `to` prop whereas `<PopoverSecondaryButton>` does

Some documentation suggestions
- A bit tangential but it would be great if Thumbprint documentation for `<Button>` accepts a ref that gets forwarded to the inner DOM element. It's unclear whether such components accept `ref` at all or what element or component instance it is going to be set to. 
- Would it be helpful to clarify that `ref` provided to the render prop of `props.launcher` is of type function (it's a callback ref as dictated by ReactPopper interface, not object ref)? Just for clarification sake and also in case where we want to compose refs to the launcher element in case we want to add more functionalities (other than Popper positioning) based on ref.

lavelle:
Thanks for the questions. Hopefully we can integrate this into the existing uses in website quite easily.

1. Click-outside-to-close is not supported. The nature of these popovers is to require the user to engage with them, for example to go to the next step of a product walkthrough. Click outside could cause accidental dismissal.

2. Width is not configurable for now. We want the contents of these to be relatively short and simple. If people need much more complex or larger UIs, they should reconsider the component. We could add a second width if we saw a valid use case thoguh.

3. We're going off the designs. Right now the idea is that the secondary button can contain a "read more" link, but the primary button should always take you to the next popover, or dismiss the popover. Again, we can easily add this if we see a valid use case.

Good suggestions for the docs. We can make those changes.

PR #342, opened 2019-06-21
Add ErrorBoundry to components that use Image

tomgenoni:


tomgenoni:
Problem was with `IntersectionObserver`

PR #347, opened 2019-06-24
Upgrade react hooks and fix warning

### LINKED ISSUE #344: Add react hooks linting
This can catch some common errors when using hooks

https://www.npmjs.com/package/eslint-plugin-react-hooks#installation

lavelle:
Fixes #344

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -18,6 +18,10 @@ module.exports = {
         // Disabled because some packages contain subcomponents. Our documentation system currently
         // requires that they be defined in one file.
         'react/no-multi-comp': 'off',
+
+        // React hooks
+        'react-hooks/rules-of-hooks': 'error',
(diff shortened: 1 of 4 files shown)


.eslintrc.js
```diff
@@ -18,6 +18,10 @@ module.exports = {
         // Disabled because some packages contain subcomponents. Our documentation system currently
         // requires that they be defined in one file.
         'react/no-multi-comp': 'off',
+
+        // React hooks
+        'react-hooks/rules-of-hooks': 'error',
+        'react-hooks/exhaustive-deps': 'warn',
```
danoc:
```suggestion
        'react-hooks/exhaustive-deps': 'error',
```

Let's do error for both? I'd rather fail the build and add `eslint-disable` lines (with comments) rather than having warnings in the logs that are easy to miss and don't include context.

PR #355, opened 2019-06-27
Refactor Tooltip into a function component

### LINKED ISSUE #354: Refactor Tooltip to use hooks
Tooltip would be a good one to do so that we can remove the `EscapeableTooltip` wrapper made in #351.

lavelle:
## This change includes:

- Refactoring Tooltip to be a function component, using hooks for state and effects
- Remove the intermediary EscapableTooltip (Fixes #354)
- Update all the unit tests to avoid using `setState()` and `state()` in favour of simulating clicks and asserting the DOM directly. This is compatible with hooks, and is a better design of black-box testing
- Split out server tests and run them in Jest's server environment
- Upgrade Enzyme (only a minor version bump) to support new Hooks testing features (see https://github.com/airbnb/enzyme/issues/2073)

## Still to do:

- [x] Fix 4 remaining failing tests


danoc:
Actually, test logs are still showing `act` errors:
https://travis-ci.org/thumbtack/thumbprint/builds/551938007?utm_source=github_status&utm_medium=notification

packages/thumbprint-react/components/Tooltip/test.jsx
```diff
@@ -552,82 +252,458 @@ describe('touch devices', () => {
...
+
+            button.simulate('mouseleave');
+            act(() => {
+                jest.runAllTimers();
+            });
+            wrapper.update();
+            expect(wrapper.find('[role="tooltip"]').exists()).toBe(false);
+        });
+
+        test('focus and blur open and close the tooltip', () => {
```
danoc: Why remove the backticks? Makes it easier to know that `focus` and `blur` are events but "open" and "close" are verbs.

lavelle: Couple of reasons:

- Makes it harder to Cmd+f for the test name to jump to the failing test after seeing it in the terminal
- Breaks terminal commands that include the test name. For example running

  ```
  yarn test:jest test.jsx --testNamePattern "`mouseleave` closes the tooltip 
  immediately if closeDelayLength is 0"
  ```

  gives the error `zsh: command not found: mouseleave`.

PR #361, opened 2019-07-01
Update ConditionalPortal to explicitly allow and handle null contents

### LINKED ISSUE #357: Children is required in `ConditionalPortal` but usage doesn't reflect that
image

We've added a conditional within most of the places that we're using `ConditionalPortal` but `children` is required within `ConditionalPortal`.

A few options: 

1. Move the `canUseDom` check outside of `ConditionalPortal`.
2. Make `children` not required.
3. Remove `canUseDom` check since `ConditionalPortal` handles that anyways.

cc: @lavelle

lavelle:
Fixes #357

### DIFF
--- a/packages/thumbprint-react/utils/ConditionalPortal.jsx
+++ b/packages/thumbprint-react/utils/ConditionalPortal.jsx
@@ -1,4 +1,3 @@
-// eslint-disable-next-line no-unused-vars
 import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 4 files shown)


lavelle:
ready for a cheeky look!

packages/thumbprint-react/utils/ConditionalPortal.jsx
```diff
@@ -1,21 +1,37 @@
...
 export default function ConditionalPortal({ shouldDisplace, children }) {
-    return canUseDOM && shouldDisplace ? ReactDOM.createPortal(children, document.body) : children;
+    if (!children) {
+        return null;
+    }
+
+    return canUseDOM && shouldDisplace ? (
+        ReactDOM.createPortal(<React.Fragment>{children}</React.Fragment>, document.body)
+    ) : (
+        <React.Fragment>{children}</React.Fragment>
```
danoc: Is `React.Fragment` needed? I think it can be just `children`.

lavelle: the fragment is here to get the linter working. it doesn't identify it as a React component if it has no JSX. See my comment in front-end-questions

lavelle: Not ideal to modify code for the linter, but it doesn't show up in the output, and it's better than a div. The alternative is not have lint errors like the proptypes warning show up, which is what caused this in the first place.

PR #363, opened 2019-07-03
Update ModalDefault demos to use function components

lavelle:
This removes boilerplate code and makes it clearer to see what code is necessary to use these components


lavelle:
PTAL

PR #364, opened 2019-07-03
Switch checkbox demo to use functions and atomic styles

lavelle:


www/src/pages/components/checkbox/react/index.mdx
```diff
@@ -114,42 +74,24 @@ class CheckboxExample extends React.Component {
...
+            Select all cities
+        </Checkbox>
+    );
 }
 ```
 
 ## Disabled checkboxes
 
-The `isDisabled` prop visually and functionally disabled the radio button. It also visually disables the related label.
+The `isDisabled` prop visually and functionally disabled the radio button.
```
danoc: Can we revert the line-length changes? Unrelated to the PR and not consistent with the rest of the MDX files.

www/src/pages/components/checkbox/react/index.mdx
```diff
@@ -15,97 +15,57 @@ Checkboxes can be checked, unchecked, or in an indeterminate state.
...
-    render() {
-        const { isChecked } = this.state;
-
-        return (
-            <Checkbox isChecked={isChecked} onChange={this.onChange}>
-                Send me promotional emails
-            </Checkbox>
-        );
-    }
+function CheckboxExample() {
```
danoc: I think it'd be good to also include a single example (possibly right after this one?) that shows how to do it with classes. We're still using classes heavily, so I can see it being confusing to folks if we _only_ document with hooks.

(The code examples do look a million times better with hooks though!)

lavelle: done

PR #367, opened 2019-07-03
Fix bug in `use-focus-trap` that caused it to refocus every render

### LINKED ISSUE #366: ModalDefault is broken in v6.3.1
This is probably due to my refactor of focus trap and hooks. When the user types a single character, the text field loses focus again.

lavelle:
Fixes #366


lavelle:
Decided to just get rid of the customisable options for now, since they weren't being used. Seems like having a ref inside an object like `focusTrapOptions` triggers a re-render. maybe it only does a shallow equality check, since `element` itself doesn't cause a re-render

lavelle:
Probably good to write an integration test for the failing case too. I'll do that on Thursday. It would probably go something like

1. Open modal with an input in it
2. Simulate pressing tab to focus the input
3. Simulate pressing a key like "a" a few times
4. Assert that the input value is the correct string eg `aaa`

This would ensure that the input is not getting unfocused after the first "a"

- [ ] Write integration test

lavelle:
PTAL

PR #369, opened 2019-07-04
Remove default fade in on Image

### LINKED ISSUE #368: Remove fade-in on the Image component
Currently the image fades in over 300ms. There was some concern about perceived loading so the plan is to remove this for now. Can re-add this as an optional prop if we provide guidelines.

tomgenoni:
Removes fade-in.

- [x] Check `Image`, `Avatar`, and `ServiceCard` deploy in FF, Safari, Chrome, Edge 18.

Fixes #368

### DIFF
--- a/packages/thumbprint-react/components/Image/index.jsx
+++ b/packages/thumbprint-react/components/Image/index.jsx
@@ -139,7 +139,7 @@ const Image = forwardRef((props, outerRef) => {
 
         aspectRatioBoxProps.style = {
             paddingTop: `${(h / w) * 100}%`,
-            overflow: 'hidden', // Prevents alt text from taking up space before `src` is populated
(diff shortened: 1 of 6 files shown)


packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -138,7 +139,7 @@ const Image = forwardRef((props, outerRef) => {
 
         aspectRatioBoxProps.style = {
             paddingTop: `${(h / w) * 100}%`,
-            overflow: 'hidden', // Prevents alt text from taking up space before `src` is populated
+            overflow: 'hidden', // Prevents alt text fvsrom taking up space before `src` is populated
```
```suggestion
            overflow: 'hidden', // Prevents alt text from taking up space before `src` is populated
```

PR #373, opened 2019-07-11
Create iOS Tokens documentation.

danoc:
This changes files in:

* `packages/thumbprint-tokens` — The changes allow us to share helper functions across the tokens package and the documentation. These functions are `formatId`, `formatValue`, and `formatSectionName`.
* `www/src` — Add iOS to the side-nav, top-nav on `/tokens`, and change tokens related JSX files to support iOS tokens.

None of these changes affect the tokens "dist" output. (If they did, at least one `*.snap` files would have changed.)

***

@tomgenoni as primary reviewer, @kgudgeirsson and @kevinmbeaulieu for feedback on the page.

***

Here is the preview:
https://deploy-preview-373--thumbprint.netlify.com/tokens/ios/


danoc:
> I'll let Tom review the actual code changes here, but the format lgtm! Only comment is that the "Install" instruction references `yarn` which I'm assuming was web-specific, so would be good to update that (maybe the "Install" line could be like `pod 'ThumbprintTokens', '~> 8.1'` and then an "Import" line below that with `import ThumbprintTokens`)

Whoops. Totally missed that! I'll fix.

danoc:
Just fixed the install and import statements:

image

danoc:
@tomgenoni — This is ready for review.

PR #389, opened 2019-07-18
Add example of importing a single SCSS file

danoc:


packages/thumbprint-scss/README.md
```diff
@@ -42,6 +42,12 @@ For projects that want to compile Thumbprint CSS components along with other SCS
 @import '[node_modules path]/@thumbtack/thumbprint-scss/components
 ```
 
+You can import a single component (`button.css`, for example) with:
+
+```scss
+@import '[node_modules path]/@thumbtack/thumbprint-scss/button
```
webbower: Is `[node_modules path]/` necessary? I see a lot of uses without it that just start with `@thumbtack`

tomgenoni: It depends on if the project it's being used in is setup to resolve to `node_modules`. https://github.com/sass/node-sass/issues/2445

danoc: I've added more info on this.

danoc:
This is ready for another look.

PR #399, opened 2019-07-25
Fix modal-default usage docs to use hooks everywhere

mateo42:
Change from this.closeModal to calling the appropriate hook.


danoc:
Thanks, Matt!

PR #402, opened 2019-07-29
Horizontal rule

dbausher:
Add new horizontal rule component to thumbprint.


packages/thumbprint-react/components/HorizontalRule/index.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+import styles from './index.module.scss';
+
+const propTypes = {
+    /**
+     * Renders the horizontal rule with the thumbtack styles.
+     */
+    lineStyle: PropTypes.oneOf(['original', 'dotted', 'dashed']),
```
mateo42: To keep the options consistent can 'original' be changed to 'solid'?

packages/thumbprint-react/components/HorizontalRule/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+    color: PropTypes.oneOf(['gray', 'gray-300']),
```
mateo42: We're translating border color to color, should this component allow arbitrary atomic classes and document that https://thumbprint.design/atomic/#section-border-colors can be used to set the color? @danoc

packages/thumbprint-react/components/HorizontalRule/index.jsx
danoc: Agreed here.

packages/thumbprint-react/components/HorizontalRule/index.jsx
danoc: Would rather not accept arbitrary colors since there are some colors we wouldn't want to support. Jon K. wouldn't want purple horizontal rules, for example.

@mateo42 — Do you think `theme` would be better?

mateo42: Depends on how you want color/themes to work in Thumbprint.  Theme implies there's a set of colors that this goes with, but color seems like its arbitrary.  I don't have a strong opinion.

danoc: Cool, let's leave as-is.

PR #421, opened 2019-08-09
Update DatePicker to have the active month be externally controlled

### LINKED ISSUE #418: Expose `month` prop in DatePicker
The currently displayed month in `DatePicker` is uncontrolled, so it's hard for the consumer to trigger a change in `DatePicker` to display a particular month (use case: user currently sees September but clicks "Today" button and we wish to display "August" in date picker); changing selectedDate doesn't change the internal state of what month is being shown. 

`react-day-picker` seems to have `month` prop (https://react-day-picker.js.org/api/DayPicker#month; in addition to `initialMonth` we are using) that makes the current month being shown fully controlled. Is it okay if thumbprint `DatePicker` also pass through `month` prop to the underlying `react-day-picker`?

lavelle:
Fixes #418

### DIFF
--- a/packages/thumbprint-react/components/DatePicker/index.jsx
+++ b/packages/thumbprint-react/components/DatePicker/index.jsx
@@ -21,6 +21,7 @@ export default function DatePicker(props) {
         value,
         onChange,
         disabledDays,
+        month,
         lastMonth,
         allowMultiSelection,
         onMonthChange,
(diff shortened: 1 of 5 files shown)


lavelle:
PTAL

packages/thumbprint-react/components/DatePicker/index.jsx
```diff
@@ -46,7 +47,8 @@ export default function DatePicker(props) {
                 disabledDays={disabledDays}
                 fromMonth={get(disabledDays, 'before', null)}
                 toMonth={lastMonth}
-                initialMonth={selectedDays[0]}
+                month={month || selectedDays[0]}
```
danoc: Do we need `selectedDays` here? Would be simpler to just make both `month` and `onMonthChange` required?

lavelle: It reduces boilerplate for the consumer. This way I can initialise `month` as `undefined`, and it will default to showing the month for the first date in `value`. Making it required means that I have to pass in the initial value twice, effectively.

danoc: Ah, I see what you mean. Sounds good!

PR #429, opened 2019-08-21
New image props to set "sizes" and disable lazy loading

### LINKED ISSUE #426: Image component: add prop to specify sizes and bypass lazyloading
We could use a prop that allows us to specify the sizes attribute manually for server rendering.  This would also bypass lazy loading as it would force the picture tag to render with sources on the server.

Use case: landing pages have above the fold hero images which cannot render until the client JS has executed.  This causes some performance metrics like Speed Index to regress due to the delay before the image is loaded.

Example:
<Image ... forceRenderWithSizes="100%">

tomgenoni:
Note the `sizes` example is intentionally blurry. Not sure how best to show this behavior.

Fixes #426


tomgenoni:
@danoc `disableLazyLoading` example working locally but not in preview deploy. SRR related?

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -61,7 +67,7 @@ const Image = forwardRef((props, outerRef) => {
 
     // `shouldLoadImage` becomes `true` when the lazy-loading functionality decides that we should
     // load the image.
-    const [inViewRef, shouldLoadImage] = useInView({
+    const [inViewRef, isInView] = useInView({
```
mateo42: when rendering without a window context will isInView will always always return false?

danoc: Yup!

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -261,6 +270,16 @@ Image.propTypes = {
      * The placeholder prevents the browser scroll from jumping when the image is lazy-loaded.
      */
     containerAspectRatio: PropTypes.number,
+    /**
+     * Overrides the default calculation of the `sizes` attribute. Generally only useful for
+     * images that are server-side rendered. See https://mzl.la/2Hh6neO for allowable values.
```
```suggestion
     * images that must be loaded before JavaScript is parsed and executed client-side. This 
     * is often paired with the `disableLazyLoading` prop. 
     * [See allowable values](https://mzl.la/2Hh6neO).
```

Changing the wording a bit to clarify "images that are server-side rendered".

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -261,6 +270,16 @@ Image.propTypes = {
...
+     */
+    forceSizes: PropTypes.string,
+    /**
+     * Disables default lazy-loading behavior. Used in rare situations where images that appear
+     * above the fold need to load instantly or are server-side rendered.
```
```suggestion
     * above the fold need to load before JavaScript is parsed and executed client-side. This
     * is often paired with the `forceSizes` prop.
```

tomgenoni:
@mateo42 @danoc This is ready for another look. The props have been combined into a single `forceEarlyRender` prop that takes an optional value that hardcodes the `sizes` attribute.

mateo42:
Curious, why did you decide to go back to one prop?

tomgenoni:
Because disabling the lazy loading behavior often requires setting a `sizes` value (other than the `100vw` case) and we couldn't think of a situation where you'd be able to use them separately. This also optimizes a bit more toward SSR.

www/src/pages/components/image/react/index.mdx
```diff
@@ -92,6 +93,44 @@ This example uses the `height` prop to "crop" the image. By default, the cropped
 />
 ```
 
+## Forcing early render
+
+In some cases, like server-side rendering, you want the HTML to be complete without the need for client-side javascript. The `forceEarlyRender` prop overrides the default `Image` behavior that would both calculate the `sizes` attribute and lazy load the image.
```
danoc:
```suggestion
The `Image` component behavior uses JavaScript to lazy load the image and and calculate the `sizes` attribute. While this is beneficial in most cases, it also means that the image will not start downloading until the client-side JavaScript is parsed and the image is scrolled into view. The `forceEarlyRender` prop allows developers to turn off lazy loading and the `sizes` calculation so that the image can load as soon as possible. This is useful for "above-the-fold" images with a predictable width in server-side-rendered environments.
```

Clarifying that it's not something you _need_ in server-side rendering. Only if it's "above-the-fold" and the width is predictable.

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -261,6 +273,11 @@ Image.propTypes = {
      * The placeholder prevents the browser scroll from jumping when the image is lazy-loaded.
      */
     containerAspectRatio: PropTypes.number,
+    /**
+     * Disables lazyloading and overrides the default calculation of the `sizes` attribute.
```
```suggestion
     * Disables lazy-loading and overrides the default calculation of the `sizes` attribute.
```

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -261,6 +273,11 @@ Image.propTypes = {
...
+     * Primarily for use in server-side rendering. See https://mzl.la/2Hh6neO for allowable values.
```
```suggestion
     * Primarily for important images in a server-side rendered environment that must be 
     * loaded before JavaScript is parsed and executed on the client. The value gets used
     * as the `sizes` attribute. [See allowable values](https://mzl.la/2Hh6neO).
```

PR #434, opened 2019-08-23
Remove noscript from image when forceEarlyRender is set

mateo42:
When forceEarlylRender is set, we don't need a <noscript> image tag since we will always render an image tag.


danoc:
Thanks for the PR, Matt!

Issue #433, opened 2019-08-22
[@thumbtack/thumbprint-react] Add ability to constrain `<ModalDefault>` height


jichan-thumbtack:
In some modal flows, I'm finding the need to constrain the modal height on desktop so that it's consistent even across modal content (whose length varies) changes.
(https://www.figma.com/file/IcbYHDSlj6A1gXJTpvM1UQF5/%F0%9F%97%93-Calendar-%2B-Availability?node-id=4188%3A1097)
I've found this to be a relatively common use case (i.e. it's a bit jarring when the modal height suddenly expands/constrains as we switch from loading state or when some collapsed content within modal gets revealed, etc.)

Currently, there doesn't seem to be a good way for consumer to apply a fixed height to modal. 
It's possible to render a child of `<ModalDefault>` as `<div>` with some fixed height, but it's deficient because
(1) We'd want the entire modal (i.e. including sticky footer and close button row) height to be consistent, not just the children
(2) When children are long, this wrapper `<div>` overflows; however, there already is a container that actually controls the overflow (`<div className="index-module__contents">`) and leads to weird styling issues (i.e. overflowing children will bleed into the bottom padding of `<div className="index-module__contents">`)

I also attempted using `<ModalDefaultAnimatedWrapper>` but `<ModalFooter isSticky>` throws a ref error (`TypeError: Cannot read property 'current' of null`) and it's just a lot of reinventing `<ModalDefault>` mechanics when the only thing I want more is height constraint. 

I would prefer if `<ModalDefault>` can accept an additional `height` prop (like `width`) that simply constrains height of the entire modal (whether it's pixel, percentage, or discrete values) and I would be happy to contribute to the change. What are your thoughts?

tomgenoni:
@jichan-thumbtack Are you not able to constrain the height of the content inside of the modal? Eg. (excuse the pseudo code):

```
<Modal>
  <ModalDefaultContent>
     <my-content-wrap height="500px">
        <my-content>i'm the content that has overflow, flexes</my-content>
        <my-footer>button, no flex</my-footer>
     </my-content-wrap>
  </ModalDefaultContent>
<Modal>
```

jichan-thumbtack:
Yeah those issues are not present with the particular example, but
(1) is a problem when you use `<ModalDefaultFooter isSticky>`. When the footer is sticky, it is placed outside of the `contents` node. 
[...]

PR #444, opened 2019-08-27
Add heightAboveSmall prop to ModalDefault (#433)

jichan-thumbtack:


jichan-thumbtack:
I added a commit to fix the bottom padding issue in Firefox and Edge

Tested on Chrome, Safari, Firefox, Edge, IE11 but please take a second look

jichan-thumbtack:
Thanks for catching that; I think now that we have a nested `<div>` for adding padding to curtain, if the click originates from there, this check fails (https://github.com/thumbtack/thumbprint/blob/master/packages/thumbprint-react/components/ModalCurtain/index.jsx#L105)

Would it be okay to duplicate (or move) this https://github.com/thumbtack/thumbprint/blob/master/packages/thumbprint-react/components/ModalDefault/index.jsx#L59 to `curtainPadding` node? curtainPadding node is now the real `target` where the click originates from.

danoc:
@jichan-thumbtack — It's been a while since I've worked in this file, but I think that will work!

danoc:
Thanks for the change! Taking a look now.

jichan-thumbtack:
I made a few tweaks to fix some issues that I saw on IE11 and Safari in the latest commit as well as some typos.
I spent literally hours attempting to fix the tall sticky footer modal on IE11 (which exists on current version as well; seems related to https://github.com/philipwalton/flexbugs/issues/216) in a trial-and-error fashion; fixing it one way for one browser breaks another browser and I thought it's too much time sink for what it's worth

danoc:
> I made a few tweaks to fix some issues that I saw on IE11 and Safari in the latest commit as well as some typos.
> I spent literally hours attempting to fix the tall sticky footer modal on IE11 (which exists on current version as well; seems related to [philipwalton/flexbugs#216](https://github.com/philipwalton/flexbugs/issues/216)) in a trial-and-error fashion; fixing it one way for one browser breaks another browser and I thought it's too much time sink for what it's worth

Thanks for digging into it! I agree that it's probably not worth it. 👍

PR #454, opened 2019-09-11
Add iconRight prop to TextButton.

### LINKED ISSUE #436: Add `iconRight` prop to `Button` and `Link` components
We've decided to support icons on the right side of the text for both the `Button` and `Link` components. This task is to add an `iconRight` prop anywhere that we currently use `iconLeft`.

Context (Thumbtack employees only):
https://thumbtack.slack.com/archives/CD01CJT4G/p1565206776041500

lavelle:


packages/thumbprint-react/components/UIAction/plain.jsx
```diff
@@ -9,17 +9,28 @@ import styles from './plain.module.scss';
...
+    if (props.iconLeft) {
+        children = (
+            <span className={styles.flexCenter}>
+                {props.iconLeft}
+                {props.children && <span className={styles.textContainer}>{props.children}</span>}
+            </span>
+        );
+    }
+
+    if (props.iconRight) {
```
webbower: Should `iconLeft` and `iconRight` be mutually exclusive props? If so, a note about that should probably be added to the docs for the components

webbower:
Also, should probably at least add tests for the change(s) you add, including tests for all permutations of `iconLeft` and `iconRight` props

lavelle:
- Fixed the mutually exclusive issue
- Add test
- Fixed docs
- Update changelog.

Not gonna add to any other components. want to keep this change simple.

PR #459, opened 2019-09-18
Replace slow `scrollparent` NPM package with faster script.

### LINKED ISSUE #416: Speed up mounting of `Image` component
The `Image` component is breaking a test in `website` because it's taking too long to mount. 

We should:

1. Do a small performance audit of it to identify which parts are the slowest when rendering it with `mount` in Enzyme.
2. Depending on the results, speed up the component.

danoc:
Fixes #416.


danoc:
@webbower — Can you take a look at this when you get a chance?

packages/thumbprint-react/components/Image/get-scroll-parent.test.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+test('defaults to `document.body`', () => {
+    const wrapper = mount(<div>goose</div>);
+    expect(getScrollParent(wrapper.getDOMNode())).toBe(document.body);
+});
+
+test('looks for parents with `overflow: scroll`', () => {
+    const wrapper = mount(
+        <div>
+            <div style={{ overflow: 'scroll' }} data-test-id="parent">
```
webbower: Is `data-test-id` an internal TUI convention that happens to differ from the public `data-test` API I added to a bunch of TUI components?

packages/thumbprint-react/components/Image/get-scroll-parent.test.jsx
```diff
@@ -0,0 +1,56 @@
...
+                </div>
+            </div>
+        </div>,
+    );
+
+    const child = wrapper.find('[data-test-id="child"]').getDOMNode();
+    const parent = wrapper.find('[data-test-id="parent"]').getDOMNode();
+
+    expect(getScrollParent(child)).toBe(parent);
+});
```
webbower: All the tests expect the immediate parent to the scroll parent and only test for the `overflow` CSS property. Recommend adding tests where the scroll parent is multiple levels above the target element and add tests for `overflow-x` and `overflow-y`

packages/thumbprint-react/components/Image/index.jsx
```diff
@@ -61,7 +61,7 @@ const Image = forwardRef((props, outerRef) => {
     // React Waypoint and lazysizes use. The React Waypoint README explains this concept well:
     // https://git.io/fj00H
 
-    const parent = canUseDOM && scrollparent(containerRef);
+    const parent = canUseDOM && containerRef && scrollparent(containerRef);
```
webbower: Was this an existing bug you found during this change?

danoc: The third-party scrollparent library returned undefined when it received null/undefined nodes. This one does not, but I prefer this behavior.

packages/thumbprint-react/components/Image/get-scroll-parent.test.jsx
danoc: Nope. That was by accident. Fixed!

packages/thumbprint-react/components/Image/get-scroll-parent.test.jsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import getScrollParent from './get-scroll-parent';
+
+test('defaults to `document.body`', () => {
```
webbower: Maybe add one more test to ensure it works when `document.body` isn't the immediate parent or switch this test to go up more than one level?

danoc:
Done! Will merge once the CI passes.

PR #462, opened 2019-09-20
Add Typescript and convert ButtonRow

lavelle:
# Tasks

- [x] Install tools and libraries
- [x] Get ESLint working
- [x] Get Jest working
- [x] Get type-checking working
- [x] Convert a single trivial component (`ButtonRow`)
- [x] Get Gatsby docs to build
- [x] Get Rollup builds working
- [x] Add type declaration files to Rollup `dist/es` and `dist/cjs` directories
- [x] Release a beta version of `thumbprint-react` and test inside Website
- [x] Get Docgen props API documentation fully working


lavelle:
Trying again with the minimum necessary changes.

packages/thumbprint-react/components/ButtonRow/__snapshots__/test.tsx.snap
```diff
@@ -3,8 +3,6 @@
 exports[`adds \`dataTest\` prop 1`] = `
 <ButtonRow
   dataTest="Duck"
-  isStackedBelowSmall={false}
```
danoc: Any idea why these get removed from the snapshot?

packages/thumbprint-react/package.json
```diff
@@ -9,7 +9,11 @@
     ],
     "sideEffects": false,
     "scripts": {
-        "prepublishOnly": "../../scripts/should-build-package.js -- rollup -c=./rollup.config.js"
+        "generateTypeDeclarations": "tsc --project ./tsconfig.types.json",
+        "buildJsFiles": "../../scripts/should-build-package.js -- rollup -c=./rollup.config.js",
+        "clear": "rm -rf dist",
+        "build": "yarn clear && yarn buildJsFiles && yarn generateTypeDeclarations && yarn generateTypeDeclarations --outDir ./dist/cjs",
+        "prepublishOnly": "yarn build"
```
danoc: Would `../../scripts/should-build-package.js -- yarn build` work? That would help speed up `yarn start` (and other scripts) when the React package doesn't change.

packages/thumbprint-react/rollup.config.js
```diff
@@ -1,20 +1,26 @@
 import babel from 'rollup-plugin-babel';
 import path from 'path';
 import copy from 'rollup-plugin-cpy';
+import resolve from 'rollup-plugin-node-resolve';
```
danoc: What changed that made this necessary?

packages/thumbprint-react/tsconfig.types.json
```diff
@@ -0,0 +1,20 @@
...
+        "rootDir": ".",
+        "declaration": true,
+        "declarationMap": true,
+        "isolatedModules": false,
+        "noEmit": false,
+        "allowJs": false,
+        "emitDeclarationOnly": true
+    },
+    // TODO(giles): figure out how to include more than this one file here.
+    "files": ["components/ButtonRow/index.tsx"]
```
danoc: If this were `index.js` instead, would TypeScript know to follow the imports and create the type definitions for `ButtonRow`?

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+        "declaration": false,
+        "allowJs": true,
+        "allowSyntheticDefaultImports": true,
+        "jsx": "preserve",
+        "lib": ["dom", "es2017"],
+        "module": "esnext",
+        "target": "esnext",
+        "moduleResolution": "node",
+        "noEmit": true,
+        "noUnusedLocals": true,
```
danoc: If this is true, then is the defined ESLint rule still needed?

tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+        "noUnusedParameters": true,
+        "preserveConstEnums": true,
+        "removeComments": false,
+        "skipLibCheck": true,
+        "noImplicitAny": false,
```
danoc: Why not set to true? Does the whole repo explode with errors?

typings/index.d.ts
```diff
@@ -0,0 +1,10 @@
+// Silence errors from libraries with no type defintions provided
+// TODO: add types to tokens and remove this line
+declare module '@thumbtack/thumbprint-tokens';
+declare module 'intersection-observer';
```
danoc: I'm surprised the two polyfills don't have definitions on DefinitelyTyped.

package.json
```diff
@@ -8,20 +8,32 @@
...
+        "test:lint:js": "eslint . --ignore-path .gitignore --ext .jsx,.js,.tsx,.ts",
+        "test:typecheck": "tsc --project tsconfig.json"
     },
     "devDependencies": {
         "@babel/core": "^7.1.2",
         "@babel/plugin-syntax-dynamic-import": "^7.2.0",
         "@babel/preset-env": "^7.1.0",
         "@babel/preset-react": "^7.0.0",
+        "@babel/preset-typescript": "^7.6.0",
+        "@thumbtack/changelog-lint": "^0.1.4",
```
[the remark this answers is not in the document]
lavelle: Done

package.json
```diff
@@ -53,6 +67,11 @@
         "prop-types": "^15.7.2",
         "react": "^16.8.6",
         "react-dom": "^16.8.6",
+        "rollup-plugin-node-resolve": "^5.2.0",
+        "rollup-plugin-typescript": "^1.0.1",
+        "tslib": "^1.10.0",
```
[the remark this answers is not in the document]
lavelle: It's not imported directly, but it's used by some of the typescript tooling to provide runtime libraries.

packages/thumbprint-react/package.json
lavelle: good idea, i can move that here

packages/thumbprint-react/rollup.config.js
lavelle: i don't remember. some quirk of babel and typescript having different module resolution logic

.eslintrc.js
```diff
@@ -1,11 +1,29 @@
 module.exports = {
-    parser: 'babel-eslint',
+    parser: '@typescript-eslint/parser',
+    parserOptions: {
+        ecmaFeatures: {
+            jsx: true,
+        },
+        // project: ['./tsconfig.json'],
```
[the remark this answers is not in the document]
lavelle: yeah, i was getting a weird error when i specified this, and i think it's not needed because this is the default filename

.eslintrc.js
```diff
@@ -22,8 +40,32 @@ module.exports = {
...
+            {
+                extensions: ['.jsx', '.tsx'],
+            },
+        ],
+        // Upgrade from warning to error
+        '@typescript-eslint/no-unused-vars': 'error',
+        '@typescript-eslint/explicit-function-return-type': 'error',
+
+        // Disable for now, eventually we should go back and enable
+        '@typescript-eslint/no-explicit-any': 'off',
```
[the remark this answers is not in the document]
lavelle: this was from the other PR where I converted a bunch of components, but it's not needed just for ButtonRow.

.eslintrc.js
```diff
@@ -22,8 +40,32 @@ module.exports = {
...
+
+        // TODO: don't add this
+        'react/button-has-type': 'off',
```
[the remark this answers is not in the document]
lavelle: this was from the other PR where I converted a bunch of components, but it's not needed just for ButtonRow.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules/typescript/lib"
```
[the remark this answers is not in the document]
lavelle: No, by default it uses a global installed typescript, if available

packages/thumbprint-react/components/ButtonRow/__snapshots__/test.tsx.snap
lavelle: Not sure. I need to look into that. I would expect them to still be there because they're now in the default function arguments

packages/thumbprint-react/tsconfig.types.json
lavelle: I tried that and it didn't seem to work.

The correct way to get all the TS files here would be to do `"includes": ["**/*.{ts,tsx}"]` or similar, since `includes` takes a glob and `files` doesn't. But for some reason when I specified that it wasn't finding any files. Maybe my glob was wrong.

tsconfig.json
lavelle: Probably not. Still not totally sure about the overlapping responsibilities of ESLint and TS. I guess initially we can leave all the rules enabled, but in the long run we can probably make ESLint faster by disabling the rules that are already caught by type-checking

tsconfig.json
lavelle: Yeah, there were a ton of errors. We will need to work on this gradually.

typings/index.d.ts
lavelle: Yeah, not sure why these two don't have them

lavelle:
Thanks for the review! Docgen seems to be fully working now. Not sure why all the info wasn't showing up before.

I pushed up a trivial fix for the `gatsby develop` issue. Gonna merge and release the new version of `thumbprint-react`, and try and convert a couple more simple components as a followup PR

PR #467, opened 2019-09-24
Convert LoaderDots and List to TypeScript

lavelle:


package.json
```diff
@@ -73,8 +73,7 @@
         "stylelint-config-recommended-scss": "^3.3.0",
         "stylelint-scss": "^3.11.0",
         "tslib": "^1.10.0",
-        "typescript": "3.6.2",
-        "typescript-plugin-css-modules": "^1.3.1"
```
lavelle: Still digging into the weird quirks with VSCode showing CSS modules errors. I don't think this is actually needed, since we don't ever process CSS modules in this repo – they're shipped as-is in the `dist` directories. So all we need is the wildcard module declaration in `index.d.ts` so that the TS compiler and VSCode know what they are. I think this plugin was conflicting with that.

danoc: Sounds good, thanks for the context!

PR #468, opened 2019-09-24
Newsletter for Sept 24

tomgenoni:


danoc:
@tomgenoni — I just added my stuff. I removed the survey results part because I'd like to first send out the more detailed version by email.

PR #470, opened 2019-09-25
Convert Tooltip to TypeScript

lavelle:


packages/thumbprint-react/components/Tooltip/test.tsx
```diff
@@ -222,7 +279,17 @@ describe('Tooltip', () => {
         expect(wrapper.find('[role="tooltip"]').exists()).toBe(true);
 
         act(() => {
-            document.dispatchEvent(new KeyboardEvent('keyup', { keyCode: ESC_KEY }));
```
danoc: Was this just never working? I'm not sure what's changed.

lavelle: It was working, but `keyCode` is a deprecated property and so not available in Typescript's default definitions for DOM APIs, so it was throwing an error. That's why I had to type-cast with `as KeyboardEventInit` to get it to not error on the old property. I added `key` and `code` too, just so that the non-deprecated properties *are* there, even though enzyme doesn't need them. Tried to put all that in the comment.

PR #472, opened 2019-09-26
Convert Image to TypeScript

lavelle:


packages/thumbprint-react/components/Image/index.tsx
```diff
@@ -19,23 +18,87 @@ import styles from './index.module.scss';
...
+    style?: {
+        objectFit?: 'cover' | 'contain';
+        objectPosition?: 'top' | 'center' | 'bottom' | 'left' | 'right';
+        fontFamily?: string;
+    };
+};
+
+type AspectRatioBoxPropsType = {
+    style?: {
+        paddingTop?: string;
```
danoc: Should this be `React.CSSProperties["paddingTop"]` or is it best to keep the type narrower like this?

packages/thumbprint-react/components/Image/index.tsx
```diff
@@ -44,12 +107,12 @@ const Image = forwardRef((props, outerRef) => {
...
     // calculated client-side because we don't know the viewport width.
 
-    const computeSizes = () =>
+    const computeSizes = (): string =>
         containerRef && containerRef.clientWidth ? `${containerRef.clientWidth}px` : '0px';
 
     // If `forceEarlyRender` is supplied use that value, otherwise use the computed width.
 
-    const sizes = forceEarlyRender || computeSizes();
+    const sizes = forceEarlyRender ? computeSizes() : '';
```
danoc: Is this supposed to be:

```suggestion
    const sizes = forceEarlyRender ? forceEarlyRender : computeSizes();
```

danoc: The issue here may be that `forceEarlyRender` can be either a boolean or a string. 

In the old code, if `forceEarlyRender` were `true`, the `sizes` attribute would be equal to `true`, not a number. This is invalid, of course, so TypeScript complains.

@tomgenoni — Should `forceEarlyRender` have been `PropTypes.oneOfType([PropTypes.bool, PropTypes.string])`? Or just `PropTypes.string`?

packages/thumbprint-react/components/Image/index.tsx
```diff
@@ -19,23 +18,87 @@ import styles from './index.module.scss';
...
+     * The placeholder prevents the browser scroll from jumping when the image is lazy-loaded.
+     */
+    containerAspectRatio?: number;
+    /**
+     * Disables lazy-loading and overrides the default calculation of the `sizes` attribute.
+     * Primarily for important images in a server-side rendered environment that must be
+     * loaded before JavaScript is parsed and executed on the client. The value gets used
+     * as the `sizes` attribute. [See allowable values](https://mzl.la/2Hh6neO).
+     */
+    forceEarlyRender?: boolean | string;
```
danoc: We could be more explicit here about the type of string allowed.

```suggestion
    forceEarlyRender?: boolean | React.ImgHTMLAttributes<HTMLImageElement>["sizes"];
```

packages/thumbprint-react/components/Image/index.tsx
lavelle: i didn't know about that type. good suggestion

packages/thumbprint-react/components/Image/index.tsx
lavelle: ah yeah, i've misunderstood this. i'm assuming `forceEarlyRender` should always be a string, and if it's a truthy string you use that, otherwise you use `computeSizes`.

packages/thumbprint-react/components/Image/index.tsx
lavelle: done

packages/thumbprint-react/components/Image/index.tsx
```diff
@@ -187,7 +250,10 @@ const Image = forwardRef((props, outerRef) => {
 
                     // `outerRef` is the potential forwarded `ref` passed in from a consumer.
                     if (outerRef) {
-                        outerRef(el);
+                        // Not all refs are callable functions, so only try and call it if it is
+                        if (typeof outerRef === 'function') {
```
danoc: Do we need the conditional on L253 if we have this?

lavelle: we do not

PR #479, opened 2019-09-30
Convert Title and Text to TypeScript

lavelle:


packages/thumbprint-react/components/Type/index.tsx
```diff
@@ -56,51 +54,43 @@ export function Title({ children, size, className, headingLevel, dataTest, id })
...
      * the ["Type" tokens](/tokens/#type) from Thumbprint Tokens or working with a designer to
      * use one of the styles on this page.
      */
-    className: PropTypes.string,
+    className?: string;
     /**
      * Name of the DOM element to render. Defaults to `p` for a paragraph tag.
      */
-    elementName: PropTypes.string,
+    elementName?: string;
```
danoc: How about `keyof React.ReactHTML`?

See `React.createElement` type definition for reference:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L219-L222

lavelle: nice find. done

PR #481, opened 2019-10-02
Convert ServiceCard components to TypeScript

lavelle:
Ready for review, just needs to rebase against #472 to fix the one type error


packages/thumbprint-react/components/ServiceCard/index.tsx
```diff
@@ -1,10 +1,35 @@
...
+     * density and the width of the rendered image. Images must have an aspect ratio of 8:5.
+     */
+    sources?: ImageSource[];
+    /**
+     * Image alt tag that's passed to `aria-label` for better accessibility.
+     */
+    alt?: string;
+}
+
+const ServiceCardImage = forwardRef<HTMLElement, ServiceCardImagePropTypes>((props, outerRef) => {
```
dcapo: Should the anonymous function be typed?

packages/thumbprint-react/components/ServiceCard/test.tsx
```diff
@@ -2,14 +2,17 @@ import React from 'react';
 import { mount } from 'enzyme';
 import ServiceCard, { ServiceCardImage, ServiceCardTitle, ServiceCardDescription } from './index';
 
-jest.mock('react-intersection-observer', () => ({
-    InView: ({ children }) => children({ inView: true, ref: null }),
-    useInView: () => [() => {}, true],
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
dcapo: Jest doesn't have types for this stuff?

packages/thumbprint-react/components/ServiceCard/index.tsx
lavelle: React generates them automatically using the `forwardRef` generic arguments `<HTMLElement, ServiceCardImagePropTypes>`. the types for `props` comes from `ServiceCardImagePropTypes` and the types for outerRef comes from `HTMLElement`

packages/thumbprint-react/components/ServiceCard/test.tsx
lavelle: it does, but here we are mocking out with dummy values that do not match the actual type signature, so it's best to ignore it.

dcapo: I've seen a bunch of people recommend avoiding `any` if at all possible. This seems like a case where it's maybe reasonable, since it's a mock. Would it be overkill to avoid using `any`?

lavelle: It's possible to avoid any here, but the resulting types are very hard to read. You're basically mocking out a real library with fake values so it's lots of function returning `void`, `null`, etc. Anyway, I did it, so you can take a look and see if you prefer this, or the `any`  stuff.

PR #482, opened 2019-10-02
Convert DatePicker to TypeScript

lavelle:


lavelle:
Oof this is a bigun. Will probably need to be a patch change since I changed the API in some (non-breaking) ways

packages/thumbprint-react/components/DatePicker/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+            <DayPicker
+                disabledDays={disabledDays}
+                fromMonth={get(disabledDays, 'before', null)}
+                toMonth={lastMonth}
+                month={month || selectedDays[0]}
+                initialMonth={month || selectedDays[0]}
+                selectedDays={selectedDays}
+                onMonthChange={onMonthChange}
+                modifiers={modifiers}
+                onDayClick={(day, { selected, disabled }): void => {
```
dcapo: Should the parameters be typed here?

packages/thumbprint-react/components/DatePicker/types.d.ts
```diff
@@ -0,0 +1,53 @@
+export type DateIsh = Date | string | number;
```
dcapo: What's the guidance around inlining types or creating a new file?

packages/thumbprint-react/components/DatePicker/index.tsx
lavelle: generated automatically by the react-day-picker library, since it knows what it should expect for this prop

packages/thumbprint-react/components/DatePicker/types.d.ts
lavelle: I've been inlining if they're only used in one place, moving to a new file as soon as they are used in two places. In this case they're needed by the main index file and the utilities.

dcapo: Makes sense

PR #484, opened 2019-10-03
Convert Carousel to TypeScript

lavelle:


packages/thumbprint-react/components/Carousel/__snapshots__/test.tsx.snap
```diff
@@ -4,8 +4,6 @@ exports[`Carousel \`selectedIndex\` supports decimal \`selectedIndex\` values 1`
 <Carousel
   onSelectedIndexChange={[Function]}
   selectedIndex={2.3}
-  spacing="0px"
```
dcapo: I'd expect the typing to have had no effect on the the test snapshots. What changed here?

packages/thumbprint-react/components/Carousel/base-carousel.tsx
```diff
@@ -4,23 +4,46 @@ import rotate from 'rotate-array';
...
     const [prevSelectedIndex, setPrevSelectedIndex] = useState(selectedIndex);
     const [isAnimating, setIsAnimating] = useState(false);
     const [isSuspensed, setIsSuspensed] = useState(false);
 
-    function reorder(items) {
+    function reorder(items: number[]): number[] {
         // The `prevSelectedIndex` doesn't update until the animation is done, so we want to
         // use that ordering until the animation is complete.
-        return rotate(items, -1 * Math.floor(prevSelectedIndex));
+        return rotate<number>(items, -1 * Math.floor(prevSelectedIndex));
```
dcapo: Noob question: what does the \<number\> syntax mean?

packages/thumbprint-react/components/Carousel/__snapshots__/test.tsx.snap
lavelle: this has happened with all of them. not sure if it's just a quirk of jest, but it seems like assigning default values via `defaultProps` vs in the function arguments affects whether they appear in the output. the default values all still work fine in real code though, so i'm assuming this is just a quirk of the snapshots.

packages/thumbprint-react/components/Carousel/base-carousel.tsx
lavelle: `rotate` is a generic function that can rotate an `Array` of any type of value. In this case it's an `Array<number>` (array of numbers) so we call `rotate<number>()` so it knows it's rotating numbers. That means it also knows the return type - in this case also `Array<number>`.

dcapo: Ack, thanks

PR #487, opened 2019-10-07
Avatar size fix

### LINKED ISSUE #486: [@thumbtack/thumbprint-react]: Avatar doesn't pass through a number on the size prop
### Issue

When using a number for the size prop the `` height is set to 0 and the Avatar is invisible. 

```
<EntityAvatar imageUrl="https://www.placecage.com/640/480" size={100} />
```

results in 

```
<img ... height="0"... />
```

which is a result of line 94 and 165:

```
height={typeof size === 'string' ? dimensions[size] : '0'}
```

### Fix

Assuming that the use case for `size` accepting a number was to support Avatar dimensions other than the 5 sizes, we need to pass through that value to the height. If it is `undefined` the height will not match the width if the intrinsic image aspect ratio is not 1:1.

tomgenoni:
This properly passes through the `size` value if is a number to allow for arbitrary image sizes. Fixes #486

cc @toraora


tomgenoni:
@danoc Should we include an example of this usage in the docs? Or should we deprecate this and try to remove it? I'm guessing this was for legacy purposes.

danoc:
Sweet, thanks for fixing. Seems like this got messed up in the TypeScript conversion.

I'd vote to not change the documentation since this is a feature that we don't want to encourage. It was added to help with a difficult breaking change a few months ago. We're unsure if we'll ever want to use it in the future and haven't scoped how hard it would be to remove all uses of size as a number.

PR #490, opened 2019-10-09
Convert Select component to TypeScript

lavelle:


packages/thumbprint-react/components/Select/index.tsx
```diff
@@ -23,136 +28,120 @@ const getUIState = ({ isDisabled, hasError }) => {
...
      */
-    isFullWidth: PropTypes.bool,
+    isFullWidth?: boolean;
     /**
      * The `value` of the `<option>` that is selected. See [React documentation
      * on `<select>` and controlled components](https://facebook.github.io/react/docs/forms.html#the-select-tag)
      * to learn more.
      */
-    value: PropTypes.string.isRequired,
+    value?: string;
```
dcapo: Should this be nullable? The old PropTypes def lists it as required.

lavelle: it should be required, good catch.

PR #499, opened 2019-10-15
Bump z-index so icon is above input

### LINKED ISSUE #498: [@thumbtack/thumbprint-react]: Input overlaps some icons
### Issue

When using Chrome form autocomplete the input background is a light blue and overlaps some medium sized icons. 

image

The container is being pulled back `-4px` and results in the overlap.
https://github.com/thumbtack/thumbprint/blob/master/packages/thumbprint-react/components/Input/index.module.scss#L199

### Possible fixes

1. Removing that should fix the immediate problem. Reducing the padding inside the input would fix up the visual spacing. 
2. Change the z-index so the icon container remains on top. 

cc: @zongThumbtack

tomgenoni:
The icon container and the input are siblings and each previously had a `z-index: 2` which means the input will overlap the icon container due to source order. By bumping the icon container to `3` it will allow wide medium icons to not be obscured by auto fill functionality that sets a background color on the input. 

This is how it will render now when input background has a color:
Screen Shot 2019-10-15 at 3 10 19 PM

Fixes #498


tomgenoni:
@danoc Looks like you'll need to re-review.

danoc:
lgtm. feel free to admin merge the changelog entry merge conflicts!

PR #503, opened 2019-10-16
Add TypeScript definitions to tokens

### LINKED ISSUE #485: Support types for tokens
It seems there are a few issues here

There is a typo in our package.json so consumers can't reference the ES exports directly. It should be `"module": "dist/javascript-es/index.js"` not `"module": "dist/javascript-esm/index.js"`. Notice the `esm`.

TypeScript will not resolve packages from the `module` field if available, so it's using the CJS file at `main`. Further reading: https://github.com/Microsoft/TypeScript/issues/21423

Generating the types is kind of a pain. I think that ideally we would use `tsc` directly to generate the `index.d.ts` file. However, the `index.js` file itself is dynamically generated by the build script from the token JSON files. So it would probably be best to run `tsc` on the contents of the `dist` directory as a second step, after running the build script. The alternative is using the existing script to also generate the `index.d.ts` file which feels sort of hacky and error prone. It would be better to use `tsc` because then the build will fail if there's a typo.

cc @danoc what do you think about this? Having autocomplete for the values of `tokens` and an error if you use an undefined token would be a big win. Right now the type of `tokens` is `any`.

lavelle:
Fixes #485 

I basically have the token generator script only generate a `typescript` output, and then use the typescript compiler to generate the CJS and ESM builds, along with their types

### Also

- Fix typo in `package.json` `module` field.
- Fix test suite to work when run from anywhere


danoc:
Just did a deploy with a clear cache. Let's see if that fixed it. 
image

packages/thumbprint-tokens/src/__snapshots__/index.test.js.snap
```diff
@@ -56,7 +56,9 @@ exports[`outputted Android matches snapshot 1`] = `
 `;
 
 exports[`outputted CommonJS matches snapshot 1`] = `
-"exports.tpBorderRadiusBase = \\"4px\\";
+"\\"use strict\\";
+Object.defineProperty(exports, \\"__esModule\\", { value: true });
```
danoc: Why does `__esModule` end up being `true`?

packages/thumbprint-tokens/package.json
```diff
@@ -22,6 +28,7 @@
         "fs-extra": "^8.1.0",
         "glob": "^7.1.2",
         "handlebars": "^4.2.0",
+        "jest": "^24.9.0",
```
danoc: Can you remove this (and the `test` script) since we run Jest at the root level?

yarn.lock
```diff
@@ -1486,6 +1486,14 @@
   resolved "https://registry.yarnpkg.com/@thumbtack/thumbprint-icons/-/thumbprint-icons-2.15.1.tgz#a655f7893479a8cffc9d8c8052b29edcd85221c2"
   integrity sha512-jXrKp9+EyLsCJcdzbEjofyZXbyjEika7TIAB+eJEwrzZCw2w2AW3gVg7aGfSy4USqX61KJ/YxZL43jps6tjdXA==
 
+"@thumbtack/thumbprint-tokens@8.3.4-beta.0":
```
danoc: Revert changes to this file?

lavelle: done

packages/thumbprint-tokens/package.json
lavelle: done

lavelle:
Okay, figured out the issue. It wasn't a cache problem. That file was still being imported, by a `.mdx` file. I didn't realise MDX files could `import` stuff. It's kind of a shame that MDX files don't have the same tooling as TS/JS files, would be nice to be able to catch errors like missing imports and have linting inside those files.

packages/thumbprint-tokens/src/__snapshots__/index.test.js.snap
danoc: Sweet, thanks for sharing!

danoc:
> Okay, figured out the issue. It wasn't a cache problem. That file was still being imported, by a `.mdx` file. I didn't realise MDX files could `import` stuff. It's kind of a shame that MDX files don't have the same tooling as TS/JS files, would be nice to be able to catch errors like missing imports and have linting inside those files.

Yeah, agreed. Wish the TypeScript support were better. Thanks for digging into this.

PR #516, opened 2019-10-18
Icon input spacing

### LINKED ISSUE #498: [@thumbtack/thumbprint-react]: Input overlaps some icons

tomgenoni:
Fixes #498


tomgenoni:
cc: @zongThumbtack I'll be browser testing this change but I don't expect any problems.

tomgenoni:
Tested on FF, Safari, iOS Safari Chrome, IE 11. Visually the inputs with icons remain identical.

danoc:
I have no idea why I didn't do it like this from the very beginning—although hopefully we don't find out why. 😅

PR #520, opened 2019-10-19
Make iframe height configurable and add border

danoc:
There's a bug in Netlify CMS right now where editing the height in Firefox doesn't work. I left a comment in their repo to see if I can help fix. https://github.com/netlify/netlify-cms/issues/1709


danoc:
@tomgenoni Can you take a look at this one?

PR #521, opened 2019-10-21
Add new React Alert component

volanStudio:
A conversion of the SCSS component `Alert` to React (ts). This work includes:

- Creation of a new React component written in typescript
- Jest tests for the component
- Addition of a `.mdx` file in the components section of the thumbprint site

Things discussed but not executed in this PR:

- Max-width for content inside the alert


www/src/pages/components/alert/react/index.mdx
```diff
@@ -92,7 +92,11 @@ export const pageQuery = graphql`
         packageTable: thumbprintComponent(name: { eq: "@thumbtack/thumbprint-react" }) {
             ...PackageTableFragment
         }
+        # Get component props by path to component file.
```
volanStudio: This seems like it should append the props to the bottom of the page, but it doesn't 🤔. I likely need some help on this to push it over the finish line.

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+        })}
+        data-test={dataTest}
+    >
+        <div className={styles.text}>{children}</div>
+    </div>
+);
+
+interface InPagePropTypes {
+    /**
+     * Text within the alertt - expects jsx.
```
danoc:
```suggestion
     * Text within the alert - expects jsx.
```

packages/thumbprint-react/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Changed
+
+-   [Patch] Convert `Alert` component to React (TypeScript).
```
```suggestion
-   [Minor] Convert `Alert` component to React (TypeScript).
```

This would be "Minor" in SemVer since it's the addition of new functionality.

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    />
+);
+
+const GoodNewsIcon = (): JSX.Element => (
```
danoc: I've been putting all the icons here so that can be used by other components within the package:

https://github.com/thumbtack/thumbprint/tree/master/packages/thumbprint-react/icons

(Although I don't remember if there actually are any that are re-used in which case that was probably a bad idea.)

packages/thumbprint-react/components/Alert/test.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <BannerAlert messageType="note">
+                <p>Alert content</p>
+            </BannerAlert>,
+        );
+        expect(
+            wrapper
+                .find('div')
+                .at(1)
+                .text(),
+        ).toBe('Alert content');
```
danoc: I recommend re-writing adding a `data-test-id="text"` attribute to the `p` element and then writing the test as `wrapper.find('[data-test-id="text"]).text()` rather than doing the `find('div')` and `at(1)`. This will make the test more resilient to changes in the component markup.

packages/thumbprint-react/components/Alert/test.tsx
```diff
@@ -0,0 +1,158 @@
...
+            <BannerAlert messageType="good">
+                <p>Alert content</p>
+            </BannerAlert>,
+        );
+        expect(
+            wrapper
+                .find('div')
+                .at(0)
+                .hasClass('goodNews'),
+        ).toBe(true);
```
danoc: Personally I think the snapshot tests are good enough for this kind of test. @lavelle, curious to know what you think.

packages/thumbprint-react/components/Alert/test.tsx
volanStudio: On it! Updated in this commit https://github.com/thumbtack/thumbprint/pull/521/commits/0737da6df1b651e9fd2b8349778f5ea432ddad10

packages/thumbprint-react/components/Alert/index.module.scss
```diff
@@ -0,0 +1,66 @@
+@import '~@thumbtack/thumbprint-tokens/dist/scss/_index';
+
+.tpAlert {
```
lavelle: By convention we call this `.root`

packages/thumbprint-react/components/Alert/index.module.scss
```diff
@@ -0,0 +1,66 @@
...
+    display: flex;
+    font-size: $tp-font__body__2__size;
+    line-height: $tp-font__body__2__line-height;
+    text-align: left;
+    padding: 12px $tp-space__3;
```
lavelle: Reason the 12px is non-standard?

packages/thumbprint-react/components/Alert/index.module.scss
```diff
@@ -0,0 +1,66 @@
...
+    color: $tp-color__white;
+    background: $tp-color__gray;
+    font-weight: 700;
+    border-radius: $tp-border-radius__base;
+
+    & a {
```
lavelle: Do we need to target an element here? I would hope our reset sheet for links would handle this. How does this interact with atomic classes, which use `!important`?

packages/thumbprint-react/components/Alert/index.module.scss
```diff
@@ -0,0 +1,66 @@
...
+        color: $tp-color__white;
+        fill: $tp-color__white;
+
+        &:hover {
+            color: $tp-color__white;
+            opacity: 0.7;
+        }
+    }
+
+    &.banner {
```
lavelle: With CSS modules, there's no need to use the `&` operator, since all class names are scoped to the file that imports them. You can flatten all of this out.

packages/thumbprint-react/components/Alert/index.module.scss
```diff
@@ -0,0 +1,66 @@
...
+    .text {
+        align-self: center;
+    }
+
+    .icon {
+        fill: currentColor;
+        flex: none;
+        margin-right: 12px;
+        position: relative;
+        top: -1px;
```
lavelle: Can you make a note about why these hardcoded offsets are needed?

packages/thumbprint-react/components/Alert/index.tsx
lavelle: I would agree with putting them there now, for consistency.

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    />
+);
+
+interface BannerPropTypes {
+    /**
+     * Text within the alert - expects jsx.
```
lavelle: You can remove "expects jsx" – this is implicit in the type definition.

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+     */
+    children?: React.ReactNode;
+    /**
+     * Defines the style of the alert
+     */
+    messageType: 'good' | 'bad' | 'warning' | 'note';
+    /**
+     * A selector hook into the React component for use in automated testing environments.
```
lavelle: Small prose typo: "A selector to hook into..."

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+     */
+    dataTest?: string;
+}
+
+const BannerAlert = ({ children, messageType, dataTest }: BannerPropTypes): JSX.Element => (
```
lavelle: By convention we use `export default function BannerAlert()` here, to put everything on one line

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+    <div
+        className={classNames({
+            [styles.tpAlert]: true,
+            [styles.banner]: true,
+            [styles.goodNews]: startsWith(messageType, 'good'),
```
lavelle: Can you just name the classes the exact same as the prop values, and use `messageType === 'good'` here, rather than a fuzzy match?

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+}
+
+const ALERT_ICONS = {
+    good: <GoodNewsIcon />,
+    bad: <BadNewsIcon />,
+    warning: <WarningIcon />,
+    note: <NoteIcon />,
+};
+
+const InPageAlert = ({ children, messageType, dataTest }: InPagePropTypes): JSX.Element => (
```
lavelle: All the same comments from above apply to this component and its type definition.

packages/thumbprint-react/components/Alert/test.tsx
lavelle: Agree that any tests like this that are simply asserting the static state of the DOM after rendering are better covered by snapshots. Reserve your custom unit tests for testing dynamic, interactive behaviour, or the logic of different functions and conditionals.

packages/thumbprint-react/components/Alert/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+interface BannerPropTypes {
+    /**
+     * Text within the alert - expects jsx.
+     */
+    children?: React.ReactNode;
+    /**
+     * Defines the style of the alert
+     */
+    messageType: 'good' | 'bad' | 'warning' | 'note';
```
lavelle: We should probably call this prop `theme`, for consistency with other components that have a set of visual styles. For example, button has `theme: 'primary' | 'secondary' | 'tertiary'`

lavelle: In addition, make it optional, probably with a default value of `note`. We could also rename `note` to `info`

packages/thumbprint-react/components/Alert/index.tsx
volanStudio: Can I do that if it is exporting 2 components form this file? I mimicked the pattern from the avatar which also exports 2 components.

packages/thumbprint-react/components/Alert/index.module.scss
volanStudio: It appears we do need to target the anchor here. Without it the style gets overridden by the default.

packages/thumbprint-react/components/Alert/index.module.scss
volanStudio: It appears to be 12px to make the height of the entire alert 44px...

packages/thumbprint-react/components/Alert/index.tsx
lavelle: Sorry, if there are two it should be `export function`. I'm trying to move over all components to that over the fat arrow syntax

lavelle:
All looking good, thanks for making those changes. Clean up all the tests to use `data-test`, and remove any that can be replaced by snapshots, and we should be good to merge.

packages/thumbprint-react/components/Alert/index.tsx
volanStudio: On it!

volanStudio:
I knocked out the rest of the improvements. The only one I had questions on was the tests. I removed all of the tests pertaining to the theme, but the convention on the icons isn't to pass a `data-test`, so it still uses `find('svg').hasClass('icon')` to test that it is accepting the class. I guess this is also potentially one that could be a snapshot.

lavelle:
Looks great, thanks for making those changes! Let's merge.

lavelle:
I will wait for @danoc to remove his request for changes before merging.

PR #526, opened 2019-10-23
Fix broken API docs for forwardRef components

### LINKED ISSUE #483: Components using `forwardRef` do not show TypeScript API docs
This seems to be an unresolved issue with `gatsby-transformer-react-docgen`. 

https://github.com/gatsbyjs/gatsby/issues/16544

Until this issue is fixed or we find a workaround these docs won't work

lavelle:
Fixes #483 

Fixes missing API docs for

- Image
- UserAvatar and EntityAvatar
- ServiceCardImage

Which are three components that have been converted to TypeScript and use `forwardRef`.

Button, Link, and UIAction also use it, but are still in JS. We should be sure to specify the props correctly when converting those components, to ensure the docs don't break.


lavelle:
This is a bit hacky since the fix is in a transitive dependency, which hasn't been published yet so I had to 

1. Fork the transitive dep (`react-docgen`) and publish their unpublished fix as a new package
2. Use `patch-package` to update the import from the library we use directly (`gatsby-transformer-react-docgen`)

I can't think of a better alternative though. It works for now, and can all be removed once those two libraries get new versions out.

danoc:
Can you also make a follow up issue to eventually revert this when the actual react-docgen package is released? It'd be helpful to have that context somewhere.

lavelle:
Issue here: https://github.com/thumbtack/thumbprint/issues/527

PR #529, opened 2019-10-24
Convert Button and UIAction to TypeScript

lavelle:


lavelle:
gatsby failing with an unrelated CMS thing. Everything looks good when I run the docs locally, including `iconRight` which accidentally got deleted but I then later fixe.d

packages/thumbprint-react/components/UIAction/get-button-props.ts
```diff
@@ -1,3 +1,13 @@
+interface ButtonProps {
+    type: 'button' | 'submit';
+    onClick?: () => void;
```
danoc: Are these not supposed to include the event as a parameter? He's `onClick`, for example:

```
(event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
```

packages/thumbprint-react/components/UIAction/test.tsx
```diff
@@ -242,38 +242,20 @@ describe('Plain', () => {
...
-    test('left and right icons can be used together', () => {
-        const wrapper = mount(
-            <Plain iconLeft={<svg>Swan</svg>} iconRight={<svg>Duck</svg>}>
-                Goose
-            </Plain>,
-        );
-        expect(wrapper.find('svg')).toHaveLength(2);
-        expect(wrapper.text()).toBe('SwanGooseDuck');
-        expect(wrapper).toMatchSnapshot();
-    });
```
danoc: Why do these two go away?

packages/thumbprint-react/components/UIAction/get-button-props.ts
lavelle: yes, i wrote this code before i knew the react dom types so well. fixed here and elsewhere.

packages/thumbprint-react/components/UIAction/test.tsx
lavelle: oops, that's what you get for cherry picking a commit from 3 months ago. good catch.

PR #533, opened 2019-10-25
Add ref support to Input component.

### LINKED ISSUE #492: Add ref support on input components
For input components it could be useful to developers who want to programmatically focus or invoke events on those inputs to have access to the DOM ref.  To do this, I think Thumbprint components should use React.forwardRef to wrap the existing components and pass the forwarded ref to the internal input element.

This would need to be considered a major version change: https://reactjs.org/docs/forwarding-refs.html#note-for-component-library-maintainers

For that reason, I don't think this is a high priority issue.

A challenge in implementing this is keeping the components as class components while being able to wrap them.  To wrap a class component you must pass the ref as a prop to the inner component.  This prop is not part of the public API for the component, so it means writing a second internal component to handle the ref.

Alternatively, we could convert the class components to functional components (some may need to use hooks to accomplish this).  I had some success trying this out on the Input component, but it ran into a snag trying to test the focusing behavior as you can't spy on the internals of the functional component.

An example of a forwardRef component: 
`const Input = React.forwardRef((props, ref) => { /* ... */ return <Input ... ref={ref} /> });`

lavelle:
Fixes #492


packages/thumbprint-react/components/InputRow/__snapshots__/test.jsx.snap
```diff
@@ -212,7 +212,7 @@ exports[`works with \`Input\` and \`Button\` 1`] = `
         }
       }
     >
-      <Input
+      <ForwardRef
```
danoc: Can we add a `.displayName` property to the component to keep the display name?

https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools

PR #535, opened 2019-10-25
Convert utils to TypeScript

lavelle:
@danoc please take a look


packages/thumbprint-react/utils/use-close-on-escape.ts
```diff
@@ -6,12 +6,12 @@ const EVENT_NAME = 'keyup';
 /**
  * React Hook to register a keyboard listener to run a function when the ESC key is pressed.
  *
- * @param {Function} doClose Function to run when ESC key is pressed.
- * @param {Boolean} isActive Whether or not the key listener does anything.
+ * @param {function} doClose Function to run when ESC key is pressed.
+ * @param {boolean} isActive Whether or not the key listener does anything.
```
danoc: Does TypeScript actually use this for the types? I'm confused why `isActive` doesn't need it's type defined below.

PR #537, opened 2019-10-25
Convert ModalCurtain to TypeScript

lavelle:


packages/thumbprint-react/components/ModalCurtain/test.tsx
```diff
@@ -47,23 +47,23 @@ describe('ModalCurtain', () => {
 
         mount(
             <ModalCurtain onCloseClick={jest.fn}>
-                {args => {
+                {(args): JSX.Element => {
                     obj = args;
```
danoc: What's the change here? I don't think I understand it.

lavelle: not sure what you're asking about specifically here, i just changed it to make the type checking pass. the `children` prop of `ModalCurtain` is a function that returns `JSX.Element`, but `null` is not included in that, so I switched it to an empty `div`

danoc:
This one needs also needs the base branch to be switched to `master`. There were a few conflicts when I tried.

danoc:
Closing this one as part of a tidying up of our issue and PR backlog. Let's re-open once we can switch the base branch to `master` and rebase.

PR #562, opened 2019-11-20
Allow access for Figma plugin

tomgenoni:


danoc:
Is this one ready for review?

tomgenoni:
After testing decided this was premature. Will revisit when we're in a better position to render useful documentation in Figma.

PR #565, opened 2019-11-21
Remove `form.scss` from Thumbprint SCSS

### LINKED ISSUE #559: Remove SCSS `form.scss` classes
Just realized that this page doesn't appear in the component sidebar and also has an incorrect title and description: https://thumbprint.design/components/form/scss/

It likely broke when we made the move to MDX a while back.

Regardless, this seems like a good candidate for removal as part of The Great Alignment work (#558):

https://github.com/thumbtack/thumbprint/blob/master/packages/thumbprint-scss/scss/form.scss

@tomgenoni — Thoughts? We'd probably need to move the placeholder styles to some other file.

danoc:
This fixes #559.


tomgenoni:
Any concerns about non-website users affected by this?

danoc:
> Any concerns about non-website users affected by this?

Just made a GitHub issue in the Help Center codebase that gives them personalized upgrade instructions. It'll be an easy change. I'm not concerned about other codebases since those are rarely upgraded and the migration is straightforward.

Issue #566, opened 2019-11-22
Rename a few components for cross-platform consistency


danoc:
We will be renaming a few React components in Thumbprint as part of our Q4 "Great Alignment" initiative. This will allow us to use consistent component naming across Figma, web, iOS, and Android.

### React component renaming

* `DatePicker` becomes `Calendar`
* `Select` becomes `Dropdown`
* `ModalDefault*` components become `Modal*`
* `Textarea` becomes `TextArea`
* `Input*` components becomes `TextInput*`

### SCSS class and file renaming

* `tp-select` becomes `tp-dropdown` and `select.scss` becomes `dropdown.scss`.
* `tp-rule` becomes `tp-horizontal-rule` and `hr.scss` becomes `horizontal-rule.scss`
* `tp-loader` becomes `tp-loader-dots` and `loader.scss` becomes `loader-dots.scss`
* `input.scss` becomes `text-input.scss`
* `tp-textarea` becomes `tp-text-area` and `textarea.scss` becomes `text-area.scss`

## How will we do it?

We'll do this in a few phases to minimize the impact on our engineers.

1. Add "This component will be renamed" alerts to documentation pages for the components that will be renamed.
2. Share this list with front-end engineers.
3. Create a codemod (with tests) that will do the renames for us. (We can re-use a codemod that we have already created.)
3. A few days later, rename the components and classes:
   1. Add aliases so that the old names temporarily continue to work. 
   2. Use the new names in the documentation.
   3. Add "This component was recently renamed" notices in the documentation so that the search can find the old components.
   4. Create 301 redirects for the old URLs.
4. Run the codemod in Thumbtack's codebase. Prioritize React changes over SCSS.
5. Two weeks later, re-run the codemod to catch any new instances of the old names.
6. Remove the aliases to the old component/class names. Release as a breaking change.
7. Two weeks later, remove the "This component was recently renamed" message.

danoc:
Decided not the rename the CSS classes because it'd be high effort and low impact. We can revisit in the future if it is causing confusion. The React work is done.

PR #569, opened 2019-11-22
Write codemod for Great Alignment rename

danoc:
Also includes tests for the codemod. Relates to #566.


danoc:
@tomgenoni — This is ready for a look. The code is simple because we've previously done a similar codemod and, at the time, extracted most of it into a reusable utility.

danoc:
@tomgenoni — Had forgotten to add `InputIcon`, and `InputClearButton`.

PR #581, opened 2019-12-10
Rename component pages in `www` folder

danoc:
This renames the component page titles and paths. It also sets up redirects from the old pages to the new ones.


www/src/pages/components/text-area/scss/index.mdx
```diff
@@ -16,20 +16,20 @@ import { InlineCode } from 'components/mdx';
...
-    id="example-basic-textarea"
+    id="example-basic-text-area"
     placeholder="Tell us about your business"
 />
 ```
 
 ## Textarea with an error
 
-The `tp-textarea--bad-news` class only changes the textarea’s color. It should be used alongside an error message that helps users advance through the form.
+The `tp-textarea--bad-news` class only changes the Text Area’s color. It should be used alongside an error message that helps users advance through the form.
```
tomgenoni:
```suggestion
The `tp-textarea--bad-news` class only changes the textarea's color. It should be used alongside an error message that helps users advance through the form.
```

danoc:
@tomgenoni — Ready for another look.

PR #590, opened 2019-12-12
Work around odd behavior of `value` in Checkbox

### LINKED ISSUE #589: Checkbox HTML includes value attribute even if prop is undefined
React adds a `value` attribute (`value=""`) to `input[type="checkbox"]` even if the `value` prop is `undefined`. This prevents the default browser behavior of `value="on"` when the `value` attribute is omitted and causes unexpected bugs if a server endpoint checks for a the checkbox value rather than existence of the checkbox in the form data.

This upstream issue describes this in more detail: https://github.com/facebook/react/issues/17590

Thumbtack employees can also look at IE-545 in JIRA for more context.

danoc:
This fixes #589.

### DIFF
--- a/packages/thumbprint-react/CHANGELOG.md
+++ b/packages/thumbprint-react/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


danoc:
@dcapo — Hey, can you take another look? Had to fix a lint error in the test.

PR #600, opened 2019-12-18
Motion tokens for Tooltip and Modal

tomgenoni:
Add new duration and easing tokens to Tooltip and Modal.

- Tooltip now fades in, SCSS only additions
- Modal replaces hard-coded easing and duration values with tokens.


packages/thumbprint-react/components/Tooltip/index.module.scss
```diff
@@ -1,12 +1,23 @@
...
+}
+
 $tooltip-max-width: 235px;
 
 .childrenWrapper {
     display: inline-flex;
 }
 
 .tooltip {
+    animation: tooltip-animation $tp-duration__2;
```
danoc: Should this also have easing, or is that only for transitions?

tomgenoni: It uses the browser default if nothing's is provided but, yes, no reason not to use a token here.

PR #607, opened 2019-12-20
Motion annotation example

tomgenoni:
Adds an example image of the motion annotations available in Figma.


tomgenoni:
It's a gif of a raccoon eating popcorn on a sofa, but you were close.

PR #609, opened 2020-01-06
Change UserAvatar, EntityAvatar Empty Background Color

TTblee:
There exists a case in the Avatar Upload component where
the Avatar empty state background matches the page background.

According to designs, the Avatar empty background color should
be `gray`, not `gray-200` (which matches the page bg color).

Therefore, change the Avatar background default color to `gray`.

https://thumbtack.atlassian.net/browse/PROG-2126

image

### DIFF
--- a/packages/thumbprint-react/CHANGELOG.md
+++ b/packages/thumbprint-react/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Changed
 
 -   [Patch] Replace existing duration and easing values for SCSS/JS tokens in `Modal`.
+-   [Patch] Change `EntityAvatar`, `UserAvatar` empty background color from `gray-200` to `gray`.
 
 ## 12.0.0 - 2019-12-17
 
--- a/packages/thumbprint-react/components/Avatar/__snapshots__/test.tsx.snap
+++ b/packages/thumbprint-react/components/Avatar/__snapshots__/test.tsx.snap
@@ -144,7 +144,7 @@ exports[`adds the \`fullName\` as \`title\` text 1`] = `
       className="initialsAvatar circleAvatar"
       style={
         Object {
-          "backgroundColor": "#fafafa",
+          "backgroundColor": "#d3d4d5",
           "color": "#2f3033",
         }
       }
@@ -251,7 +251,7 @@ exports[`does not add the \`fullName\` as \`alt\` text when no image is provided
       className="initialsAvatar circleAvatar"
       style={
         Object {
-          "backgroundColor": "#fafafa",
+          "backgroundColor": "#d3d4d5",
           "color": "#2f3033",
         }
       }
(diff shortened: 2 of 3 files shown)


tomgenoni:
We might also consider leaving the default as `gray-200` but providing the option to change it in cases like this. 

@jonpkerwin you have thoughts on this?

jonpkerwin:
@tomgenoni @TTblee 

We could do a few things here. My initial thought is to make that avatar a full blank avatar, which would give it a light bg color and an intial. Would also be a good indicator of what the user would see if no avatar is uploaded.

We could also make the avatar white with a blue dotted outline, like an upload affordance. Or fill that square with a more conventional user silhouette. 

I also support the bg color being alterable!

tomgenoni:
Since this the first instance of this I'd suggest doing a one-off until @jonpkerwin can do a proper audit and design recommendation. How about...

Screen Shot 2020-01-07 at 10 28 26 AM

```
<div className="flex">
    <div class="ba b-dashed b-blue bw-2 br2">
        <EntityAvatar size="large" />
    </div>
</div>
```

mattybow:
@tomgenoni I think we would need to redesign how the page looks if we go with your suggestion 
Screen Shot 2020-01-07 at 11 56 48 AM

tomgenoni:
You could also try a gray border if the blue is too harsh. 

I'd recommend working with your designer and @jonpkerwin to work out a solution for now. We'd need to do more work on the design side so we can confident the fix is what we want for this (and possibly other situations like this) going forward.

TTblee:
Looks like the designers have chatted and we are going with the original proposal: https://thumbtack.slack.com/archives/CACQMDGQ1/p1578426931003300

@tomgenoni  @jonpkerwin 
My opinion is that it's better to have the "empty" Avatar as a same, single gray-ish color (like what currently exists). Otherwise, giving developers background color options may introduce unwanted disparity down the line.

The current case of this "empty" Avatar is only on Pro Onboarding, where the user does not have a name, so there are no initials to prefill the avatar. I don't think the flexibility of changing the background color of the avatar in this state is necessary.

I am siding with simply changing the "empty" Avatar color from gray-200 --> gray (what currently exists in this commit) to keep the change simple and simplify future design decisions, but if you think otherwise, please let me know 🙂

tomgenoni:
Ok, thank you @TTblee 

I don't personally have an objection to making it `gray` per se but we'd want to run a change like this through our component process. In this case @jonpkerwin would do an audit of usage and determine what pattern / style we want to recommend, update the Figma component(s), and then we'd update the React/iOS/Android components. It's possible (though unlikely) that it's already used against `gray` and if so we'd be right back to the same problem. 

I'd recommend chatting with your designer and Jon if there's a complication like this in the future so we can plan out the change. Happy to discuss in person if you'd like.

jonpkerwin:
Since this is only used here this is fine with me. In all other instances you would see either a filled or a blank avatar anyways.

TTblee:
Bumping this PR. If there are any considerations, I'm happy to continue the conversation. Otherwise, I would love to submit/close out this PR 🙂

tomgenoni:
Closing until @jonpkerwin can do a proper audit of this use case. We'll go with the one-off solution discussed above.

PR #612, opened 2020-01-08
Box shadow atomic class fix for IE 11 to Edge 18

### LINKED ISSUE #476: [@thumbtack/thumbprint-atomic]: Box shadows in IE 11 are too faint
### Issue
The Atomic classes for box shadows, e.g., `shadow-300`, render considerably fainter in IE 11. 

### Suggested Fix
Apply CSS to IE 11 only that applies `outline: 1px solid $tp-color__gray-300;`. 

`border` behaves similarly and looks good because it follow the contours of the border radius, but a `border` also changes the dimensions of the box, decreasing the internal width by 2px (1px on each side), and in some rare cases that could lead to visual inconsistencies. The height is also increased if it's not constrained. 

#### IE 11 Screenshot 
Screen Shot 2019-09-27 at 3 10 01 PM

cc @xanderwebs

tomgenoni:
Adds CSS targeted only at IE 11 through Edge 18 browsers with different, darker `box-shadow` values so they render closer to what modern browsers use. In the original ticket the fix was to use an `outline` but visually it's too far off.

The top row is how it renders in Chrome/Safari/Firefox. The bottom row is how it looks in IE/Edge with fix applied. 

Screen Shot 2020-01-07 at 1 51 04 PM

This was tested in all Chrome/Safari/Firefox, iOS Safari, Android Galaxy S9, Windows Phone, IE 11 and all Edge versions including Chromium 79 beta. 

Fixes #476


packages/thumbprint-atomic/__snapshots__/test.js.snap
```diff
@@ -4624,15 +4672,31 @@ exports[`compiles correctly 1`] = `
...
+  _:-ms-lang(x) {
+    box-shadow: 0 1px 8px rgba(0, 0, 0, 0.15) !important;
+  }
   .l_shadow-200 {
     box-shadow: 0 2px 4px rgba(0, 0, 0, 0.15) !important;
   }
+  .l_shadow-200,
+  _:-ms-lang(x) {
+    box-shadow: 0 1px 10px rgba(0, 0, 0, 0.2) !important;
+  }
```
danoc: Why is it that this doesn't end up darkening in Chrome/FF/Safari? I would guess that this `.l_shadow-200` would "win" over the three lines before it.

tomgenoni: According to the CSS spec "The prelude of the qualified rule is parsed as a <selector-list>. If this returns failure, the entire style rule is invalid." https://www.w3.org/TR/css-syntax-3/#style-rule

"Therefore when the other browsers try to parse the selectors, they find the _:-ms-lang(x) selector to be invalid and so ignore the entire rule (including .selector)" [Stack Overflow answer](https://stackoverflow.com/questions/53925353/how-does-the-ms-langx-fix-for-edge-and-ie-work )

But, it looks like the selector order is changing during compilation, even though that doesn't appear to matter to invalidating the rule-set. This is the source SCSS:

```
_:-ms-lang(x),
.shadow-200 {
    box-shadow: map-get($ie-edge-shadow, 200);
}
```

but the output is flipping the selectors:

```
 .shadow-200,
  _:-ms-lang(x) {
    box-shadow: 0 1px 10px rgba(0, 0, 0, 0.2) !important;
  }
```

I'll look into why it's flipping.

tomgenoni: `cssnano` is ordering the selectors with its `unique-selectors` functionality. Since the order doesn't matter, disabling that cssnano rule doesn't seem warranted. I've switched the order in the SCSS source to prevent the order mismatch.

danoc: Ok cool, good catch!

danoc: Also the description was helpful. I didn't realize it gives up on the whole rule if it doesn't understand part of the selector.

packages/thumbprint-atomic/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Added
+
+-   [Minor] Adding IE/Edge selectors to `box-shadow` to bring visual parity with modern browsers. (#476)
```
danoc:
```suggestion
-   [Patch] Adding IE/Edge selectors to `box-shadow` to bring visual parity with modern browsers. (#476)
```

Changing because it's a bug fix.

PR #615, opened 2020-01-08
Fix page header in CMS component pages

danoc:
Currently the platform header is missing from component pages that are using the CMS. This brings it back so that users can switch platforms.


danoc:
Here's an example of a page currently missing the header:
https://thumbprint.design/components/popover/usage/

PR #617, opened 2020-01-09
Remove rename notices from component docs

danoc:
This removes it from the React and the SCSS component docs.

The React ones can be removed because the migration is complete.

I never ended up actually starting the SCSS name changes. I propose that we leave the old classes as-is and revisit the rename migration for that in the future. It's high effort and low impact IMO.


danoc:
@tomgenoni — Missed TextInput. Just updated.

PR #618, opened 2020-01-09
Fix Coda deprecation warnings in gatsby develop

danoc:
This sanitizes the column names from Coda, replacing spaces with underscores. It fixes the Gatsby V3 deprecation message that was appearing on `gatsby develop`.


danoc:
I messed something up here. WIll fix.

danoc:
Actually this seems to now be correct. Will re-open.

PR #623, opened 2020-01-09
Fix the legend on the components' Overview page

bawjensen:
Frankly I didn't bother testing this, but it seems straightforward enough.

Might be worth abstracting these into global constants acting as enums...? I can bake that into this PR if we think that's a good change.


danoc:
Thanks! It works perfectly:

https://deploy-preview-623--thumbprint.netlify.com/components/overview/

I'm cool w/ leaving it as. I'm still fiddling with the Coda setup so it's possible that we'll be changing things around again in the near future.

PR #625, opened 2020-01-10
Filter out duplicate classname, hack from Atomic doc listing

### LINKED ISSUE #622: [@thumbtack/thumbprint-atomic]: IE/Edge box shadow classes showing in Atomic docs
New IE/Edge only classes were added in #622 

Screen Shot 2020-01-09 at 3 11 28 PM

Will have to find a way to black list those classes.

tomgenoni:
Fixes #622


www/src/components/thumbprint-atomic/table/index.jsx
```diff
@@ -10,45 +10,53 @@ const cleanSelector = selector =>
...
-            ))}
-        </tbody>
-    </table>
-);
+const Table = ({ atomicClasses }) => {
+    const atomicClassesHackless = atomicClasses.filter(
+        // There are `shadow-*, _:-ms-lang(x)` hack selectors in the Atomic source code that
+        // provide darker box-shadows for IE/Edge. This filters out those selectors to prevent them
+        // from rendering in these docs.
+        item => item.selectors[1] !== '_:-ms-lang(x)',
```
danoc: Probably a little better to do `item => !item.selectors.includes('_:-ms-lang(x)')`.

tomgenoni: Yeah, better. Just updated.

PR #626, opened 2020-01-13
Rendering full list of Calendar PropTypes

### LINKED ISSUE #586: Calendar component is missing props in the documentation
I'm only seeing these three props.

image

volanStudio:
Fixes #586

Oh boy, this one is weird! I had to get pretty deep in Typescript to sort it out, but as usual I learned a ton about ts and this repo.

### Problem:
Previously the Calendar component had `PropTypes` in an `index.d.ts`. That was preventing the thumbprint.design `ComponentFooter` from accessing the props and displaying them on the component's page.

### Solution:
This PR moves the `PropTypes` into the component file. Doing so necessitated a refactor of `utilities.ts`. As far as I can surmise, the reasoning for having the `PropTypes` in their own file was to allow for referencing in `utilities.ts`.

### Remaining questions:
1. Is the outcome worth adding logic into the component file?
2. Is there another way to do this that I just overlooked?
3. Should we ship this :) ?


danoc:
Thanks for the PR, Lucas!

I think another approach here is to:

1. Keep the utilities file separate 
2. Export the types from `index.tsx` with `export interface PropTypes` and import them in the utilities file

I think that would solve the problem with a small amount of change. What do you think?

volanStudio:
That was my first inclination, but I got a `cycle` error since `index.ts` imports `utilities.ts`. I think that is why it was in `types.d.ts` in the first place - they previously were imported in both `index.ts` imports `utilities.ts`.

danoc:
> That was my first inclination, but I got a `cycle` error since `index.ts` imports `utilities.ts`. I think that is why it was in `types.d.ts` in the first place - they previously were imported in both `index.ts` imports `utilities.ts`.

Ah, right. That makes sense. Let's see if @lavelle has any suggestions. Otherwise this is looking good to me.

lavelle:
Thanks for doing this! I do remember seeing something in the `react-docgen` docs about them not being able to reference types from a different file. Sorry for not catching that sooner.

Unfortunately `react-docgen` basically never throws any errors or warning messages :( it just silently returns no props if anything goes wrong.

As for this PR, it looks good overall. I think having half of the utilities in a separate file and half in the index file is confusing though. I would say lets just combine everything into the index file – the types, utility functions, and the main component. then we can just have a single test file too, which tests both the utilities and the component.

Also, make sure you change the title of this PR to something more descriptive before merging into master.

volanStudio:
Great suggestion! 

I moved utilities into the index. I still need the `types.d.ts` since we are referencing the `Modifier` type from the `react-day-picker` library and we need to append a possible `null` value.

PR #629, opened 2020-01-14
Convert Pill to TypeScript

lavelle:


typings/index.d.ts
```diff
@@ -1,6 +1,8 @@
 // Silence errors from libraries with no type defintions provided
 declare module 'intersection-observer';
 declare module 'object-fit-images';
+// TODO(giles): remove once we add types to tp-icons
+declare module '@thumbtack/thumbprint-icons';
```
danoc: Why is this needed if `icon` is `React.ReactNode`?

lavelle: it's imported by the test suite, which throws an error right now

danoc: Ah, ok. Makes sense!

danoc:
The Netlify build is failing because of a change to tokens. I'll fix that right now.

danoc:
Fix for the build issue is here:
https://github.com/thumbtack/thumbprint-tokens/pull/27

PR #634, opened 2020-01-15
Make `Image` work with new version of dep

danoc:
The `Image` component was infinitely re-rendering with `react-intersection-observer@8.24.1`. It did not have this version with `v8.23.0`, the previous version.

I think the the issue happened because of this change: https://github.com/thebuilder/react-intersection-observer/commit/b31487b0c796cf35ecd3faaaa61da743a6abe9a2

This PR changes how we we combine multiple refs into one, following a suggestion in the `react-intersection-observer` README. This fixes the infinite re-render issue, although I'm still trying to understand why.

### DIFF
--- a/packages/thumbprint-react/CHANGELOG.md
+++ b/packages/thumbprint-react/CHANGELOG.md
@@ -8,6 +8,10 @@
 -   [Patch] Rewrite `Pill` component in TypeScript.
 -   [Patch] Rewrite `InputRow` component in TypeScript.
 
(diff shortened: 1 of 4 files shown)


danoc:
@mateo42, @tomgenoni — This is the PR. Still trying to understand why the fix works though.

danoc:
Made an issue in `react-intersection-observer`: https://github.com/thebuilder/react-intersection-observer/issues/291

PR #637, opened 2020-01-21
Create AlertBanner component

TTblee:
This commit includes
- component index files
- documentation
- test
- relevant icons
- deprecating existent BannerAlert


packages/thumbprint-react/components/AlertBanner/test.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import forEach from 'lodash/forEach';
+import AlertBanner from './index';
+import { BlockedFilled, InfoFilled, WarningFilled } from '../../icons';
+
+describe.only('AlertBanner', () => {
```
danoc: Forgot the `.only`?

TTblee: fixed

danoc:
@TTblee — I made some small tweaks to the docs:

* Removed reference to "Global Header" since that's a term that only makes sense internally.
* Updated text in the examples to use realistic examples. (We're inconsistent about this in our docs, but it's something we'd like to do more of.)

LMK if those look good to you. If it does, I'll merge!

PR #638, opened 2020-01-24
Fix tokens page

danoc:
The tokens pages used to make the assumption that each iOS, Sass, and Android token would also have a JavaScript equivalent. This is no longer true.

(The JavaScript value is currently used on the iOS and Android tokens documentation pages to show color examples since JavaScript values can be rendered on the web whereas UIColor would have to be converted.)

This also renames `data` to `exampleData` so that it's more clear that the value passed in is used to render an example of the token.


danoc:
This will unblock #635.

PR #640, opened 2020-01-29
Jan 28 newsletter

tomgenoni:
https://deploy-preview-640--thumbprint.netlify.com/updates/notes/2020-01-28


danoc:
@tomgenoni — I've updated this. Made a tweak to the format so that it's grouped by project rather than platform. Makes more sense IMO since we want all the platforms to move together.

PR #654, opened 2020-03-11
Convert Modal to TypeScript

lavelle:


packages/thumbprint-react/components/Modal/components/transition.tsx
```diff
@@ -1,37 +1,39 @@
...
+    onEntered?: () => void;
+    onExited?: () => void;
+}
+
+interface StateTypes {
+    stage: Stage;
+}
+
+export default class Transition extends React.Component<PropTypes, StateTypes> {
+    currentTimeout: number | null = null;
```
danoc: Could this go in the `constructor`? I'd rather avoid adding a Babel plugin in this PR if we don't need.

packages/thumbprint-react/components/Modal/index.tsx
```diff
@@ -106,221 +112,189 @@ const ModalAnimatedWrapper = ({
...
      * Allows the React `style` prop to be passed through to the rendered element.
      */
-    style: PropTypes.shape(),
-};
-
-const modalContentFullBleedDefaultProps = {
-    className: '',
-    style: {},
-};
+    style?: object;
```
danoc: `React.CSSProperties` would be helpful here.

packages/thumbprint-react/components/Modal/index.tsx
```diff
@@ -344,6 +318,14 @@ class ModalFooter extends React.Component {
                         return <div className={styles.modalFooterFluid}>{children}</div>;
                     }
 
+                    if (stickyFooterContainerRef === null) {
+                        return <div className={styles.modalFooterFluid}>{children}</div>;
+                    }
+
+                    if (stickyFooterContainerRef.current === null) {
+                        return <div className={styles.modalFooterFluid}>{children}</div>;
+                    }
```
danoc: Combine into one conditional since the returns are the same?

packages/thumbprint-react/components/Modal/index.tsx
```diff
@@ -384,17 +367,17 @@ class Modal extends React.Component {
...
             onOpenFinish,
-            shouldCloseOnCurtainClick,
-            shouldHideCloseButton,
-            width,
-            heightAboveSmall,
+            shouldCloseOnCurtainClick = true,
+            shouldHideCloseButton = false,
+            width = 'medium',
+            heightAboveSmall = 'auto',
         } = this.props;
```
danoc: Why are the default values needed here?

packages/thumbprint-react/components/Modal/index.tsx
lavelle: done

packages/thumbprint-react/components/Modal/index.tsx
lavelle: it seems like typescript isn't quite smart enough to be sure that the value of `.current` is not null if you combine them into a more complex boolean expression. I will refactor out the shared return value though.

lavelle: actually, i guess i had the wrong boolean expression before. got it to work now

packages/thumbprint-react/components/Modal/index.tsx
lavelle: why wouldn't they be? these are the values that were previously provided by `defaultProps`

packages/thumbprint-react/components/Modal/components/transition.tsx
lavelle: I fixed this in the end by storing these things on `this.state` instead of `this`, which I think was always what we should have been doing.

If you want to put things on `this` directly, TS requires that you declare them as class properties first, but it looks like the TS-Babel plugin doesn't support that syntax directly. We haven't run across this so far because I already did most of the work to convert all the other components to function components.

A followup task is probably to convert the Modal to a function component too, but for now storing these values on `this.state` allows us to complete this TS migration without adding a new Babel plugin.

lavelle:
not sure why travis is hanging

packages/thumbprint-react/components/Modal/components/transition.tsx
danoc: Ah, interesting.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes

Based on that MDN link, would `Transition.currentTimeout = 123` outside of the classs also be a way of declaring a class property? (No need to change the code, just curious.)

packages/thumbprint-react/components/Modal/index.tsx
danoc: My bad. I got confused because `ModalAnimatedWrapper` and `Modal` have a lot of overlap. This is correct though.

packages/thumbprint-react/components/Modal/components/transition.tsx
lavelle: no, you would also need to extend the class' type definition before you could add that property to it.

PR #655, opened 2020-03-11
Convert ModalCurtain to TypeScript

lavelle:


packages/thumbprint-react/components/ModalCurtain/index.tsx
```diff
@@ -35,45 +40,38 @@ const propTypes = {
      *
      * Modals that do not have transitions will only use the `entered` and `exited` stages.
      */
-    stage: PropTypes.oneOf(['entering', 'entered', 'exiting', 'exited']),
+    stage?: 'entering' | 'entered' | 'exiting' | 'exited' | null;
```
danoc: When would it be `null`?

lavelle: We set it to `null` in the constructor of the `Transition` component. It only becomes `entering` or `exited` in `componentDidMount`.

So it should never be null while the component is mounted, but we have to account for the fact that it can be null while unmounted in the TS definition.

Not sure if that's the desired behaviour, but it is the current behaviour.

lavelle:
Please take another look at this

PR #657, opened 2020-03-12
Add React component FAB

zhaotongpan:


packages/thumbprint-react/components/Fab/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+        >
+            {children}
+        </button>
+    );
+}
+
+interface IconFabProps {
+    /**
+     * Icon from [Thumbprint Icons](/icons/) to render within the button. It must be one of the
+     * `small` icons.
```
danoc:
```suggestion
     * `medium` icons.
```

www/src/pages/components/fab/react/index.mdx
```diff
@@ -0,0 +1,64 @@
...
+import { graphql } from 'gatsby';
+import { ComponentHeader, ComponentFooter } from 'components/thumbprint-components';
+
+<ComponentHeader data={props.data} />
+
+## Text Fab
+
+```jsx
+<TextFab>
+    <ContentActionsFilterSmall />
```
danoc:
```suggestion
    <ContentActionsFilterSmall className="mr2" />
```

www/src/pages/components/fab/react/index.mdx
```diff
@@ -0,0 +1,64 @@
...
+    Filters
+</TextFab>
+```
+
+### Active Text Fab
+
+```jsx
+<TextFab active>
+    <ContentActionsFilterSmall />
```
danoc:
```suggestion
    <ContentActionsFilterSmall className="mr2" />
```

packages/thumbprint-react/components/Fab/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+     */
+    icon: React.ReactNode;
+    /**
+     * Visually highlights the fab if it‘s active.
+     */
+    active?: boolean;
```
danoc: Rather than `active` as a boolean, I think it'd be better to use `theme="primary"` and `theme="secondary"` as we put in the API review meeting notes.

A few benefits:

* Terminology is familiar since buttons also use these themes.
* If we wanted to add a third theme in the future, it'd be easy with an enum but difficult with a boolean.

packages/thumbprint-react/components/Fab/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Function that will run when the fab is clicked on.
+     */
+    onClick?: () => void;
+}
+
+export function IconFab({ icon, active = false, onClick = noop }: IconFabProps): JSX.Element {
```
danoc: To make this accessible to screen readers, we need to add a prop that adds a `aria-label` to the `<button>`. We've used a prop called `accessibilityLabel` in other components.

packages/thumbprint-react/components/Fab/index.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import classNames from 'classnames';
+import noop from 'lodash/noop';
```
danoc: We've used `noop` in some other components but have been trying to move away from it since it isn't providing anything helpful. (Plus, it makes testing a bit harder since, if I remember correctly, `noop` is truthy.)

PR #662, opened 2020-03-18
Improve component overview page

danoc:


danoc:
Here's a design we could use that doesn't rely on images.

image

egoens:
Was this an experimental branch or was this always intended to be merged into master? It's certainly stale at this point but I would be interested in knowing more about this PR.

danoc:
@egoens — It was intended to be merged but we never got around to finishing it. Here's a preview of what we had:

https://deploy-preview-662--thumbprint.netlify.app/components/overview/

IMO there's a few good things about this layout, but we'd probably want to redesign it so that it doesn't rely on creating an image for each component.

egoens:
Thanks for the clarification. I do like this approach. It reminds me of the approach I've seen on https://evergreen.segment.com/components.

It makes finding components a bit more digestible and provides a rough glimpse into the intention of each component. There is some overhead with generating an img per component, but I don't find it to be too offputting. It's possible that we could use the same image in the Figma library to reinforce the visual approach and give the visual a purpose in both spaces.

PR #667, opened 2020-03-20
Convert more www components to TypeScript

lavelle:


www/src/components/mdx/code-block/index.tsx
```diff
@@ -1,24 +1,37 @@
...
     if (language === 'jsx' && shouldRender) {
-        return <ReactCodeBlock {...props} />;
+        return <ReactCodeBlock theme={theme}>{children}</ReactCodeBlock>;
     }
 
-    /* eslint-disable react/no-danger */
     return (
         <div>
-            {language === 'html' && shouldRender && (
+            {(language as Language & 'html') === 'html' && shouldRender && (
```
danoc: What does this mean?

lavelle: chatted offline. the typecast is necessary because Prism doesn't appear to support html out of the box (although it still seems to work for us)

PR #679, opened 2020-04-07
Add cursor style atomic css category

zhaotongpan:


packages/thumbprint-atomic/src/packages/cursor.scss
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    .l_cursor-auto {
+        cursor: auto !important;
+    }
+
+    .l_cursor-pointer {
+        cursor: pointer !important;
+    }
+}
```
danoc, line 56: Are there cases where we'd want to have the cursor behave differently depending on a breakpoint? If not, I think it's best to not include these. Keeps the file size small!

zhaotongpan: I think there could be scenarios when people only want to apply pointer cursor in small screen, which means a combo of "s_cursor-default" + "cursor-pointer" would be needed

danoc: Hm, do you have examples of scenarios? I can't really think of any.

zhaotongpan: the example will depend on the design requirement. I'm proposing this because i've seen similar cases when using other atomic classes, such as "flex" + "s_db" so only small screen has flexbox. basically it's the scenarios when you want to make special style only **below** instead of **above** a breakpoint

danoc: Yeah, the responsive classes are definitely a helpful pattern. But we don't add them for all classes because it makes more sense for some than it does for others. Those layout classes are commonly used because layouts always change when designing responsive screens. 

We don't add responsive classes to colors though since it's a lot less likely that colors change depending on the browser viewport. Cursor changing would be even less likely since the cursors don't exist on touch devices.

(Forgot to respond yesterday. Sorry for the delay!)

zhaotongpan: got it. that makes sense!

packages/thumbprint-atomic/src/packages/cursor.scss
```diff
@@ -0,0 +1,24 @@
...
+    .m_cursor-pointer {
+        cursor: pointer !important;
+    }
+}
+
+@include tp-respond-above($tp-breakpoint__large) {
+    .l_cursor-pointer {
+        cursor: pointer !important;
+    }
+}
```
danoc, line 24: Wouldn't it also make sense to remove these responsive ones though?

danoc:
Just released as v4.1.0 of Atomic. Thanks!

PR #682, opened 2020-04-23
Add isReadOnly prop to TextArea react component

### LINKED ISSUE #681: [@thumbtack/thumbprint-react]: TextArea doesn't expose readonly HTML attribute
TextInput exposes this attribute, and the native HTML <textarea /> has the attribute. Seems like it should be exposed/configurable.

bawjensen:
Closes #681


danoc:
LGTM! Will merge and release once this build passes.

PR #684, opened 2020-05-26
Add form attribute to Button and TextButton

lavelle:


lavelle:
Won't add this now since it's not supported in IE 11

PR #687, opened 2020-06-09
Upgrade Prettier

danoc:
Adds changelog entries to packages that will need republishing.


danoc:
Not sure why "Upgrade Gatsby and MDX packages" appears in the commit history. Shouldn't be an issue though since the changes don't appear in the code diffs.

PR #688, opened 2020-06-09
Add native lazy-loading support

danoc:


danoc:
Sorry for the noise here. This is good for review!

packages/thumbprint-react/components/Image/index.tsx
```diff
@@ -118,36 +118,49 @@ const Image = forwardRef<HTMLElement, ImagePropTypes>((props: ImagePropTypes, ou
...
+    // Skip over the scroll parent calculation if the browser supports native lazy-loading.
+    if (!browserSupportsNativeLazyLoading) {
+        parent = canUseDOM && containerRef ? scrollparent(containerRef) : null;
+        root = parent && (parent.tagName === 'HTML' || parent.tagName === 'BODY') ? null : parent;
+    }
 
-    // `shouldLoadImage` becomes `true` when the lazy-loading functionality decides that we should
-    // load the image.
+    // `isInView` becomes `true` when the lazy-loading functionality decides that we should load
+    // the image. We'll only add the `inViewRef` to the `picture` element if the browser
```
lavelle, line 141: missing part of the sentence here?

danoc:
I'm starting to realize this won't be as impactful as I hoped.

Two issues:

* Chrome is [very eager about when it starts to load the image](https://calendar.perfplanet.com/2019/native-image-lazy-loading-in-chrome-is-way-too-eager/). There's some [discussion about improving this](https://github.com/whatwg/html/issues/5408).
* `sizes` gets calculated with JavaScript. If we defer to the browser for lazy-loading, the browser will assume the image is 100% width, often choosing the wrong size in a set of responsive images. There's [discussion about improving this as well](https://github.com/whatwg/html/issues/4654).

This CR does improve performance on pages that have heavy use of carousels. But we can fix the carousel bug separately and see performance improvements without these downsides.

PR #689, opened 2020-06-16
Improve lazy-loading of images within carousels.

danoc:
The `Image` component will now create more than one `IntersectionObserver` if the `Image` is within a scrollable container that is not the `body` or `html` elements.

This means that images within a carousel that uses `overflowX: auto` would only render if the image is visible within the carousel _and_ the carousel is visible within the `body`. Previously, the first few images in the carousel would render right away since only one `IntersectionObserver` could be created.


danoc:
Jest will pass once #690 is merged.

danoc:
This code now contains the contents of #690. I really should have built this PR on top of #690 rather than the other way around, but 🤷‍♂️. I'll admin merge once tests pass since both PRs were approved.

danoc:
@lavelle — Did cross-browser testing and found out that Edge 15 doesn't support the `isIntersecting` property. Changed the code to use `entry.intersectionRatio > 0` instead.

Can you take another look?

danoc:
Ok I've finished my second round of cross-browser testing.

PR #702, opened 2020-07-15
Add initialFocus prop to ModalCurtain

### LINKED ISSUE #685: Make `ModalCurtain` initial focus configurable
The `ModalBase` component focuses on the first focusable element when opening but `ModalCurtain` focuses on the modal container. Focusing on the container is better for screen readers since users will be able to read the content of the modal from top to bottom.

Occasionally, it may make sense to skip directly to an input. Take this modal for example:

screenshot of a modal

It's currently not possible to force `ModalCurtain` to focus on the input when the modal opens. Even if you grab the `ref` of the `TextInput` and call `.focus()`, the ModalCurtain will take away the focus as soon as its focus trap code runs.

I propose adding an `initialFocus` prop that will make this configurable.

lavelle:
Fixes #685


lavelle:
I went for a selector-based API. Interested to get your thoughts on this. Would you prefer, this, or a ref?

danoc:
@lavelle — I've revisted this since it's a requirement for moving away from `ModalBase`. I know this is originally your PR, but can you PTAL at the changes I've made?

lavelle:
LGTM! I think you will need to approve since it's my CR

Issue #705, opened 2020-07-16
Support Swift syntax highlighting


danoc:
We've had a lot of interest in adding more iOS documentation to Thumbprint. We should add syntax highlighting for `swift`.

Swift isn't included as a default language with `prism-react-renderer`, so I think we'd need to follow the instructions in the "How do I add more language highlighting support?" part of the `prism-react-renderer` FAQ: 

https://github.com/FormidableLabs/prism-react-renderer#faq

PR #707, opened 2020-07-16
Add ios docs for toast, radio, loader dots, carousel layout

DarvishKamalia:
Swift syntax highlighting is missing (#705), and I can't see headers on local (#705), but everything else should be good to go.


www/src/pages/components/carousel/ios/index.mdx
```diff
@@ -0,0 +1,30 @@
+---
+title: CarouselLayout
```
danoc, line 2:
```suggestion
title: CarouselLayout
title: Carousel
```

The page has to have the same name as the component on the other platforms otherwise the documentation isn't able to group it together:

https://deploy-preview-707--thumbprint.netlify.app/components/carousel/ios/

Would it be 👍 to name the page title "Carousel" even if the concept within the codebase is "CarouselLayout"? FWIW the search functionality would pick up both terms.

danoc:
LGTM! I'll let @kevinmbeaulieu +2 on the iOS content.

PR #710, opened 2020-07-23
Remove Netlify CMS (1/2)

danoc:
This removes the CMS but keeps the CMS generated Markdown files in their existing locations. They continue to render as before. A follow-up PR will move all of those files into `www/src/pages`.


danoc:
@kevinmbeaulieu — Just FYI ^

danoc:
Fixed lockfile conflict. Merging.

PR #711, opened 2020-07-23
Remove Netlify CMS (2/2)

danoc:
This moves files from `www/src/cms` to `www/src/pages`.

It also runs the files through Prettier.


danoc:
Fixed simple merge conflict. Will merge.

PR #717, opened 2020-07-27
Fix typo in iOS Toast docs

danoc:


danoc:
@kevinmbeaulieu — This was a test to see if the CODEOWNERS feature worked. Looks like it did!

PR #719, opened 2020-08-04
Upgrade documentation packages (2/2)

danoc:
Routine maintenance in preparation for some more devDependency package upgrades (babel and rollup).


danoc:
Not sure why TypeScript was failing suddenly if the package upgrade was in #718. 🤷

PR #722, opened 2020-09-17
Add onError prop to Image, and use initials fallback for Avatar

lavelle:


lavelle:
I think this works well. The only thing to be wary of is that if people set a new `src` inside `onError` that is also an error, they could trigger and infinite loop. Not sure what the best way to handle that is

lavelle:
I also did the alternative CR to add the logic into a wrapper in Help Center instead https://gerrit.thumbtack.io/c/help/+/152824 (only accessible to Thumbtack employees)

I'm leaning more towards that second option, because it feels like this logic for testing the validity of the URL is more the caller's responsibility

I'm interested to know your thoughts though. I think it would be nice if Image could handle this directly, but there's no clean way to do it without introducing the possibility of an infinite loop.

I guess it depends on how "smart" we want the Image to be. it already handles a lot of stuff like lazy loading and responsive sizes, not sure if fallback images should be part of that.

Here's some interesting prior art for a `react-image` component that supports a `useImage` hook with fallbacks https://www.npmjs.com/package/react-image

lavelle:
We've decided to go with the Help Center CR for now, since this is not a common use case, but we can revisit this if it comes up again and there's a strong argument for adding it to the Image API

PR #723, opened 2020-09-29
Update Android button documentation

dzirbel:
I've refactored Android buttons in https://gerrit.thumbtack.io/c/android/+/152791 to use the same class and add some properties.


dzirbel:
Prettier is complaining - any ideas for a quick fix? I just submitted the PR from the github editor, I'd rather avoid cloning and trying to run it myself, and the check doesn't list issues. But I can just run myself if it'll be a pain to fix up.

danoc (approved):
Just fixed the Prettier issue! Thanks for the change. I'll merge when CI passes.

PR #732, opened 2020-11-11
Add event argument to Dropdown methods

lavelle:


lavelle:
PTAL. Checked the docs and it looks right

image

danoc (approved):

PR #734, opened 2020-11-18
Fix types for Calendar component

lavelle:
This fixes two bugs

- The shared types were in a `.d.ts` file, and referred to non-existent values, which led to them being resolved as `any`, meaning consumers of the library had no type-safety for values they passed to the `value` prop.
- The prop validation did not properly handle `null`, meaning an error was thrown when the prop `value={null}` was passed alongside `disabledDays`


packages/thumbprint-react/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+### Fixed
+
+-   [Patch] Fix fatal error being thrown when passing `null` to the `value` prop of `Calendar`.
```
danoc, line 7: Was this actually fixed?

image

Maybe we should add a regression test?

danoc:
One question, LGTM otherwise.

danoc (approved):
We should rewrite those tests at some point to use the public APIs of the components.

Made an issue to track this: https://github.com/thumbtack/thumbprint/issues/736

danoc:
Released as 14.3.2. 🥳

lavelle:
🙇

PR #740, opened 2020-11-25
Add iOS documentation for Navigation Bar component

kevinmbeaulieu:


www/src/pages/components/navigation-bar/ios/index.mdx
```diff
@@ -0,0 +1,139 @@
...
+
+#### Shadowless
+
+![Shadowless navigation bar](/img/ios-navigation-bar-shadowless.png)
+
+#### Transparent
+
+![Transparent navigation bar](/img/ios-navigation-bar-transparent.png)
+
+### Content Style
```
omoralesvivo, line 31: Are these associated with specific appearances of do we mix and match? Feels weird to keep them as separate types/parameters if they are going to be tied up to particular appearances.

omoralesvivo (approved):
More a question than a comment but otherwise I don't see any issues with this.

www/src/pages/components/navigation-bar/ios/index.mdx
kevinmbeaulieu: We mix and match- most commonly (/only) with the `.transparent` appearance, where we might use either content style depending on whether the content underneath is light or dark

omoralesvivo: Ok. Still think it might make more sense to make the content style part of the theme, we'd only end up adding one prebuilt type after all (transparent appearance becomes transparent/light and transparent/dark).

No need to change it now and besides we probably will need to have this conversation again if we ever go for dark mode, which I'd want to tackle at some point for AX reasons mostly.

danoc (approved):
Looks great do me.

@kevinmbeaulieu — Do you have merge permissions here? If not LMK so I can set that up.

omoralesvivo (approved):
Ok looking good to me 👍🏼

Issue #743, opened 2020-12-10
TextInput component (and maybe others) don't work with `useRef`


danoc:
The `TextInput` component has two uses of `ref`:

1. The public API accepts a `ref` prop
2. The component has an internal `ref` to programmatically focus the input element

The component attempts to merge the two refs but currently doesn't work if the consumer uses `useRef`.

This internal/external ref pattern is common for UI libraries and there are a few solutions:

* https://www.davedrinks.coffee/how-do-i-use-two-react-refs/
* https://github.com/gregberge/react-merge-refs

We should probably use one of these solutions rather than trying to do it ourselves. We should also audit the existing components that expose `ref` to see if they also don't work with `useRef`.

PR #744, opened 2020-12-10
Fix `useRef` usage with `TextInput`

danoc:
Relates to #743.


danoc:
I can't figure out how to get TypeScript working with that test. Any tips?

mattybow:
@danoc seems like changing L259 to
```let ref: React.RefObject<HTMLInputElement> = { current: null };```
works

danoc:
Decided to close this because we don't have an immediate need for it. We should revisit if the problem is raised again.

PR #748, opened 2020-12-18
Write documentation for ios fab

1994ericseo:


www/src/pages/components/fab/ios/index.mdx
```diff
@@ -0,0 +1,137 @@
+---
+title: Fab
```
kevinmbeaulieu, line 2: nit: probably best to spell out "Floating Action Button" for the title

www/src/pages/components/fab/ios/index.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+Use this function to set the icon (`icon`) and the read-out accessibility label (`accessibilityLabel`) for the icon fab.
+
+#### `public var theme: Theme`
+
+#### `public override var isHighlighted: Bool`
+
+#### `public init(icon: UIImage, accessibilityLabel: String, theme: IconFloatingActionButton.Theme = .primary)`
+
+#### `public override var intrinsicContentSize: CGSize`
```
kevinmbeaulieu, line 137: Not sure if this is why the build is failing, but I think you need to add this snippet at the bottom for the nav bar to show up properly-

```suggestion
#### `public override var intrinsicContentSize: CGSize`

export const pageQuery = graphql`
    {
        # Get links to by path to display in the navbar.
        platformNav: allSitePage(filter: { path: { glob: "/components/fab/*/" } }) {
            edges {
                node {
                    ...PlatformNavFragment
                }
            }
        }
    }
`;
```

www/src/pages/components/fab/ios/index.mdx
danoc: The React one is also called "Fab" so probably better to leave this like this for now and change both in a follow-up PR if we'd like.

danoc (commented):
LGTM! Will let @kevinmbeaulieu approve.

kevinmbeaulieu (approved):

PR #762, opened 2021-03-22
Remove unnecessary canUseDOM check from Popover.

danoc:
This was causing issues in on RR pages described in:

https://thumbtack.slack.com/archives/C0GKYQVTR/p1616022352008000

(Link is internal only.)

After creating a small reproduction, I realized that removing `canUseDOM` fixes the problem. I'm not 100% sure why, but I imagine it is related to React's client-side hydration. `canUseDOM` is always false on the server but defaults to `true` on the client. I imagine this isn't an issue with components like Tooltip because they default to being closed and require user interaction to appear.

### DIFF
--- a/packages/thumbprint-react/CHANGELOG.md
+++ b/packages/thumbprint-react/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Fixed
+
+-   [Patch] Fix server-side rendering Popover bug.
+
 ## 14.3.3 - 2020-11-19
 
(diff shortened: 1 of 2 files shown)


danoc:
FWIW I wasn't able to reproduce this on our documentation site. Not sure why that is. But I believe it's a safe change since I've tested it in website and also the SSR thumbprint.design site doesn't throw any SSR related errors after the change.

mxpan (approved):

lavelle (approved):

Issue #592, opened 2019-12-13
Can't upgrade workspace packages in clean repo install


danoc:
More context:
https://github.com/yarnpkg/yarn/issues/7734#issuecomment-565558049

Downgrading Yarn may work, but I haven't tried that yet.

danoc:
If you've installed Yarn with Homebrew, this can be used to downgrade to 1.16.0:

```
brew unlink yarn
brew install https://raw.githubusercontent.com/karlhorky/homebrew-core/f8bc263f93dcfd1cc50c98976a8503c8a7cb032f/Formula/yarn.rb
brew switch yarn 1.16.0
```

danoc:
Confirmed that downgrading Yarn works.

PR #763, opened 2021-04-07
Use Yarn Policies to use a local copy of Yarn 1.19

lavelle:
This avoids the issue in https://github.com/thumbtack/thumbprint/issues/592 where newer versions of Yarn throw an error


danoc:
Great, thanks!

danoc (approved):

PR #764, opened 2021-04-08
Upgrade react-day-picker to 7.4.10 and allow React 17 peer dependency

lavelle:


packages/thumbprint-react/package.json
```diff
@@ -17,8 +17,8 @@
     },
     "peerDependencies": {
         "prop-types": ">= 15.5.7 < 16.0.0-0",
-        "react": ">= 16.3.2 < 17.0.0-0",
-        "react-dom": ">= 16.3.3 < 17.0.0-0"
+        "react": ">= 16.3.2",
+        "react-dom": ">= 16.3.3"
```
danoc, line 21: It's good to keep this since a future breaking changes in React will eventually be incompatible with our components. For example, React 18 may drop support for `UNSAFE_componentWillUpdate`.

```suggestion
        "react": ">= 16.3.2 < 18.0.0-0",
        "react-dom": ">= 16.3.3 < 18.0.0-0"
```

danoc (commented):
This made me realize that `ModalBase` is not compatible with React 17 since `componentWillUpdate` needs to be renamed to `UNSAFE_componentWillUpdate`.

This PR is a pre-requisite for this peer dep change:
https://github.com/thumbtack/thumbprint/pull/765

PR #776, opened 2021-06-21
Allow React 17 peer dependency

jtabach:


jtabach:
I ran `yarn install` from the wrong directory. There seems to be some dependency issues when adding the resolutions. I tihnk I will move the resolutions to a separate PR.

packages/thumbprint-react/package.json
```diff
@@ -63,5 +63,9 @@
         "rollup-plugin-typescript": "^1.0.1",
         "tslib": "^1.10.0",
         "typescript": "^3.9.6"
+    },
+    "resolutions": {
+        "react": "17.0.2",
+        "react-dom": "17.0.2"
```
danoc, line 69: These shouldn't be in this file because `resolutions` is something that the consumer of a package sets.

If we want to upgrade the version of React used in this codebase, we'd do it by editing the root `package.json` and `www/package.json`. But that can be done in a separate step since it's a different task than indicating that the `@thumbtack/thumbprint-react` package works with React 17.

danoc (commented):
Do we need to upgrade the version of `react-day-picker`?

jtabach:
> Do we need to upgrade the version of `react-day-picker`?

I don't think we need to necessarily, but I could if you'd like. Bumping the peerDependency is needed for upgrading `react-day-picker`.

danoc:
@jtabach — Thanks for the change. The Jest snapshot test failed. You'll need to run `yarn run test:jest -u` and commit the changes.

danoc (approved):

PR #792, opened 2022-02-16
Bump readme node version

fow5040:
Minor change - noticed this error message on setup and wanted to update the readme with the correct node version (as of now):
```
[1/5] 🔍  Validating package.json...
error @: The engine "node" is incompatible with this module. Expected version "12.x". Got "8.9.4"
error Found incompatible module.
```


egoens (approved):
Good catch. Thanks for updating.

danoc (approved):

PR #795, opened 2022-03-15
Add support for generic type to Dropdown component

lavelle:


packages/thumbprint-react/components/Dropdown/index.tsx
```diff
@@ -27,7 +27,7 @@ const getUIState = ({
     return 'default';
 };
 
-interface PropTypes {
```
danoc, line 30: This used to be just `string` but now it's `string | number`. Is number definitely a valid value type?

danoc (commented):
One question, but LGTM otherwise.

packages/thumbprint-react/components/Dropdown/index.tsx
lavelle: Yeah, input value can actually be of type `value?: string | ReadonlyArray<string> | number | undefined;`, but I figured we don't want to support the `string[]` case for now

PR #797, opened 2022-03-15
Move thumbprint-atomic and thumbprint-scss to dart-sass

lavelle:
This helps us prepare for Node 16


packages/thumbprint-atomic/generate-css.js
```diff
@@ -0,0 +1,35 @@
...
+    const processedCss = await postcss([autoprefixer, cssnano]).process(css, {
+        from: undefined,
+    });
+
+    const outputFileName = fromFile.replace('src/', 'dist/').replace('.scss', '.css');
+
+    return fse.outputFile(outputFileName, processedCss);
+};
+
+(async () => {
```
danoc, line 32: Could we add a comment here explaining why we need the file and can't use the CLI directly?

packages/thumbprint-atomic/package.json
```diff
@@ -16,7 +16,7 @@
     "scripts": {
         "start": "yarn clear && yarn generateCss",
         "clear": "rm -rf dist",
-        "generateCss": "mkdir -p dist && node-sass -q src/atomic.scss dist/atomic.css --importer=./node-sass-importer.js && postcss dist/atomic.css --map --use autoprefixer css-mqpacker cssnano --replace",
+        "generateCss": "mkdir -p dist && ./generate-css.js src/atomic.scss dist/atomic.css && postcss dist/atomic.css --map --use autoprefixer css-mqpacker cssnano --replace",
```
danoc, line 19: Looks like the `src/atomic.scss dist/atomic.css` part can be removed since `./generate-css.js` doesn't accept args?

packages/thumbprint-atomic/generate-css.js
lavelle: Done

packages/thumbprint-atomic/package.json
lavelle: Done

lavelle:
Went down a bit of a rabbit hole converting to TS but I think it's worth it because it makes sure that the importer plugin works with whatever version of Dart sass we're using

packages/thumbprint-atomic/package.json
```diff
@@ -5,6 +5,7 @@
     "author": "Tom Genoni <tom@thumbtack.com>",
     "main": "dist/atomic.css",
     "style": "dist/atomic.css",
+    "type": "module",
```
danoc, line 8: One thing that may be worth testing is whether this will cause issues with the clients that use Thumbprint Atomic. IIRC, this tells Node that the package is uses ES Module syntax. I've run into problems in the past with packages that do this (new versions of `lodash-es`, in particular) since we don't transpile `node_modules` by default.

It might be worth testing this out by adding this line to a `node_modules/@thumbtack/thumbprint-atomic/package.json` file in some project and then running the project to see if it builds.

lavelle: Yeah, the situation with Typescript / ts-node / ESM / native ES modules is a mess right now. I found all sorts of conflicting answers on GH/SO. I'll run your test and report back. I'm hoping it's ok since this is a CSS only package

lavelle: yep, added `"type": "module",` to `/Users/glavelle/Thumbtack/frontend/node_modules/@thumbtack/thumbprint-atomic/package.json` and Nextjs builds fine

danoc: Ok cool, thanks for checking!

danoc (approved):

danoc:
Thanks! I fixed the `yarn.lock` conflicts in that last commit.

PR #806, opened 2022-04-07
Improve Filterchip Component

IsaiahCollazo99:
Fixes the border width for an unselected chip (2px => 1px)

Fixes the thinning of the chip when selected

Allows adding icons


packages/thumbprint-react/components/Chip/test.tsx
```diff
@@ -80,3 +83,15 @@ test('renders the correct text when clicked', () => {
     expect(button).toBeTruthy();
     expect(button).toHaveTextContent('Remove Beverages');
 });
+
+test('renders the `icon` passed in', () => {
+    const wrapper = mount(
+        <FilterChip
+            text="Lightning"
+            icon={<ContentModifierLightningTiny />}
```
danoc, line 91: I recommend just doing `<svg />` here. It keeps the snapshot small and simplifies the test a little.

packages/thumbprint-react/components/Chip/index.tsx
```diff
@@ -34,7 +40,10 @@ export default function FilterChip({
             onClick={onClick}
             aria-pressed={isSelected}
         >
-            {text}
+            <div className="flex items-center">
```
danoc, line 43: This might seem counter-intuitive, but we actually don't use Atomic CSS classes in Thumbprint React components.

If we were to use them in Thumbprint React, we'd be creating an implicit dependency between the version of Thumbprint React and Thumbprint Atomic that a codebase is using.

For example, imagine:

1. We add a class called `sr-only` to Atomic which hides elements from everything except a screenreader.
2. Thumbprint React starts using `.sr-only`.
3. Website upgrades to the latest version of Thumbprint React but doesn't upgrade Thumbprint Atomic.

We'd then end up in a situation where the class is used in Thumbprint React but isn't loaded on the page.

Alternatively we could import Thumbprint Atomic _within_ Thumbprint React. But that causes a new problem where the Atomic classes get loaded multiple times on a page.

packages/thumbprint-react/components/Chip/index.module.scss
```diff
@@ -5,13 +5,20 @@
...
+        border: 1px solid $tp-color__gray;
         color: $tp-color__blue;
         background-color: $tp-color__white;
     }
 }
+
+.iconWrap {
+    width: 14px;
+    height: 14px;
+    margin-right: 8px;
```
danoc, line 22:
```suggestion
    margin-right: $tp-space__2;
```

danoc (commented):
Nice job! Just 3 small comments. It's looking great though.

packages/thumbprint-react/components/Chip/test.tsx
IsaiahCollazo99: Done

packages/thumbprint-react/components/Chip/index.tsx
IsaiahCollazo99: Ahh, this makes sense. Moved the styling to the scss file

packages/thumbprint-react/components/Chip/index.module.scss
IsaiahCollazo99: Done

www/src/pages/components/chip/react/index.mdx
```diff
@@ -39,6 +39,22 @@ function FilterChipExample() {
...
 ```
 
+With the `icon` prop:
+
+```jsx
+function FilterChipIconExample() {
+    const [isSelected, setIsSelected] = React.useState(true);
+    return (
+        <FilterChip
+            text="I'm Selected"
```
egoens, line 49: Small suggestion. The initial state is not selected and it feels strange to see a label that says "I'm selected" but it's not. Maybe we can use a more generic label instead?

IsaiahCollazo99: Ah, meant to change this. Thanks for pointing it out. Done.

egoens:
Very excited about this merge and nice work! The Figma design library now supports this approach https://www.figma.com/file/YrpLNWoOwvEADbyFBWzuKv0V/Thumbprint-Global?node-id=1134%3A16

Looks like we're a rebase away from getting this moving?

danoc (dismissed):
Great job! I'll merge and release once CI passes.

danoc (approved):

PR #817, opened 2022-04-29
Upgrade postcss to prep for Node 16

lavelle:
Also loosen the node engines range to allow testing with Node 16

### DIFF
--- a/packages/thumbprint-global-css/package.json
+++ b/packages/thumbprint-global-css/package.json
@@ -26,7 +26,8 @@
         "autoprefixer": "^9.0.1",
         "cssnano": "^4.1.0",
         "node-sass-tilde-importer": "2.0.0-alpha.1",
-        "postcss-cli": "^6.1.1",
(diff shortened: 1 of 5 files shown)


packages/thumbprint-global-css/package.json
```diff
@@ -26,7 +26,8 @@
         "autoprefixer": "^9.0.1",
         "cssnano": "^4.1.0",
         "node-sass-tilde-importer": "2.0.0-alpha.1",
-        "postcss-cli": "^6.1.1",
+        "postcss": "^8.4.12",
```
danoc, line 29: Why is this needed to make it work? Does the CLI have a peerDep?

danoc (approved):

packages/thumbprint-global-css/package.json
lavelle: it's not even specified as a peerDeep -- the CLI just errors if you run without it.

PR #818, opened 2022-05-04
Upgrade to use Node 16

lavelle:
After much prep work in other PRs, this PR is finally as simple as changing the version number.


dcapo (approved):

egoens:
A small change, but we should probably update the documentation to call out the requirement for 16.x now https://github.com/thumbtack/thumbprint/blob/master/CONTRIBUTING.md?plain=1#L29

PR #820, opened 2022-05-11
Convert ModalCurtain unit tests from Enzyme to RTL

lavelle:
This will allow us to more robustly test the unmount DOM cleanup behaviour in future.

### DIFF
--- a/packages/thumbprint-react/components/ModalCurtain/__snapshots__/test.tsx.snap
+++ b/packages/thumbprint-react/components/ModalCurtain/__snapshots__/test.tsx.snap
@@ -65,3 +65,24 @@ exports[`ModalCurtain renders modal in \`exiting\` stage 1`] = `
   />
 </body>
 `;
+
+exports[`ModalCurtain renders text that is passed in 1`] = `
+<body>
+  <div>
+    <div>
+      <button
+        type="button"
+      />
+    </div>
+  </div>
+  <div
+    aria-label="Modal"
+    role="dialog"
+    tabindex="-1"
+  >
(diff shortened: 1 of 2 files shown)


ankitprahladsoni (approved):

packages/thumbprint-react/components/ModalCurtain/__snapshots__/test.tsx.snap
```diff
@@ -1,341 +1,88 @@
...
-    >
-      <div
-        aria-label="goosegoosegoose"
-        role="dialog"
-        tabIndex={-1}
-      />
-    </Portal>
-  </ConditionalPortal>
-</ModalCurtain>
+<body>
```
dcapo, line 4: Why is it that these snapshots have changed so drastically? I wouldn't expect the RTL migration to change it much. Or is there something else going on here?

lavelle: I think it's just a difference in the approach of the two libraries. RTL doesn't provide this stringified representation of the react component tree AFAICT, because it's more about testing what gets rendered to the DOM rather than implementation details of the React components. So I changed the string that's getting snapshotted, from the Enzyme React tree string, to the RTL DOM contents.

I'm still new to this though, if you think that snapshotting the component tree makes more sense, and if there's a way to do that in RTL, I'm happy to do that.

I did try snapshotting the result of `wrapper.asFragment()`, but that just returned a single line `<DocumentFragment />` which is definitely not what we want.

PR #821, opened 2022-05-12
Add unit test for ModalCurtain unmount cleanup

lavelle:
So I'm pretty confused. This test passes, and the `screen.debug()` calls show what you'd expect. We can still add this unit test, but it doesn't explain why the RF tests were failing.

### DIFF
--- a/packages/thumbprint-react/components/ModalCurtain/test.tsx
+++ b/packages/thumbprint-react/components/ModalCurtain/test.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { render, screen } from '@testing-library/react';
 import ModalCurtain from './index';
+import { mount } from 'enzyme';
 
 const ESC_KEY = 27;
 const ENTER_KEY = 13;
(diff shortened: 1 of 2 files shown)


dcapo:
> So I'm pretty confused. This test passes, and the `screen.debug()` calls show what you'd expect. We can still add this unit test, but it doesn't explain why the RF tests were failing.

Would this same test fail if run from website-legacy? How do we test against the behavior of the bug that's occurring in the RF? Seems like the test needs to fail first in order for it to be a worthwhile regression test.

lavelle:
So @ankitprahladsoni looked into this more this morning. It seems like the necessary conditions for the failure are

1. A component that uses createPortal somewhere
2. A test suite that mixes Enzyme and RTL
3. Not calling .unmount() at the end of the Enzyme tests

Only the RF met all three of this criteria, which is why it was happening there. I was able to make it happen in this PR by doing the same.

The interesting quirk though is that even after adding the logic from the RTL docs to have the Modal manually remove the document node when unmounted, the test suites still fail. I think this is probably because Enzyme is a leaky abstraction, and doesn't properly run the full React lifecycle. At this point, we can basically say that this is an Enzyme problem, and it's better to add .unmount() where needed, than to modify our components to appease it.

@ankitprahladsoni is modifying the RF tests to use unmount. I'll close this PR as won't do. Thanks for the help, team.

PR #823, opened 2022-05-31
Add additional event handler props to Link and ThemedLink

lavelle:
This allows them to be used inside a Tooltip


packages/thumbprint-react/components/Link/index.tsx
```diff
@@ -107,6 +127,27 @@ export interface LinkProps {
      * Function to fire when clicking on the anchor. This should be used alongside the `to` prop.
      */
     onClick?: () => void;
+    /**
+     * Function that runs when the user hovers on the link.
+     */
+    onMouseEnter?: () => void;
+    /**
+     * Function that runs when the user hovers on the link. Unlike `onMouseEnter`, `onMouseOver`
```
zargold, line 135: for accessibility: would be good to remind people that these should be used carefully and covered by equivalent strategies. For example a popover that appears only upon mouseEnter must also be openable by focus. Not sure the exact convention we use here, but I've seen some component libraries use console.warn when using mouseEnter without onFocus.

lavelle: Yeah, this is mostly for use inside our Tooltip component, which automatically passes all 5 of these props in. This comment is consistent with the one for the Button component, but I can follow up and update both if we can think of better wording.

PR #827, opened 2022-06-08
Upgrade to Yarn 3

lavelle:


danoc:
Not sure why this didn't fail, but we have some places where we use `--frozen-lockfile`:

https://github.com/thumbtack/thumbprint/blob/e56a8baef0999924cb52fe67a1830cd7177e42ac/.github/workflows/test.yml

lavelle:
Ah good catch. It worked because that flag is only deprecated, not removed.

danoc:
Ah, interesting. Looks like it was removed in 2.0.0 but then added back as deprecated to make the upgrade path easier.

danoc (approved):

PR #834, opened 2022-07-05
Add a `sr-only` class to Atomic

danoc:
This will allow us to add elements to the page that will only be visible to screen readers.

The CSS properties and values come from Tailwind:
https://tailwindcss.com/docs/screen-readers


egoens (approved):
LGTM! Next step: provide example use cases. 👏🏼

PR #839, opened 2022-08-01
Adds documentation for accessible color usage

egoens:
Provides guidance on how to apply accessible colors with the existing color values provided by Thumbprint.

https://thumbtack.atlassian.net/browse/DS-1024


egoens:
@danoc This PR includes the representation of the 500 values, which are not represented currently in the live environment. The inclusion is to help support the accessibility documentation that is the intention behind this PR.

egoens:
This PR is also in a draft state due to the solicitation of feedback from the design accessibility group.

danoc (dismissed):

danoc:
PR looks great to me. Feel free to merge once it looks good to the design accessibility group.

egoens:
@danoc Looking [at the preview](https://deploy-preview-839--thumbprint.netlify.app/guide/product/color/), this is a good example of why we might want to consider the 600 level to the atomic classes as well.

egoens:
@danoc Adding a bit more color to the conversation, the feedback from the design team was that the images were a bit complex and a bit to complex to describe in alt text in some cases. Instead of using imagery to describe accessible pairing, the approach was to use HTML instead.

Based on the current approach, this would mean we'd need to include the 500 & 600-level atomic classes for our color values or update the approach to use the tokens to add the appropriate color in the documentation.

danoc (approved):

PR #841, opened 2022-08-10
Add support for InputRow to Dropdown

lavelle:


lavelle:
image

danoc (approved):

egoens:
This is great @lavelle ! We may want to consider setting the `border-top-right-radius` and `border-bottom-right-radius` to smooth out the transition between the select and button element.

image

lavelle:
hmm, i have set that value, not sure why that's not working

PR #844, opened 2022-08-29
Add `dataTestId` to Modal components

danoc:


packages/thumbprint-react/components/Modal/index.tsx
```diff
@@ -79,7 +84,7 @@ const ModalAnimatedWrapper = ({
                                 [styles.curtainInnerShouldPageScrollAboveSmall]: shouldPageScrollAboveSmall,
                             })}
                             onClick={shouldCloseOnCurtainClick ? curtainOnClick : undefined}
-                            data-test="thumbprint-modal-curtain"
+                            data-testid="thumbprint-modal-curtain"
```
lavelle, line 87: I guess this isn't technically a major version bump since this dataTest prop is not part of the API right now and is only used by Thumbprint's own test suite. Have you done a quick grep in the website to make sure nothing is depending on this anyway. Wouldn't want to do a seemingly safe version bump and break tests.

lavelle (dismissed):

egoens:
@danoc it looks like the changelog has conflicts. If we can get that aligned, I'll get this merged.

danoc:
@egoens — I'm going to close this one because Giles raised a good point (I think it should be a major version bump) and I don't actually need the change anymore.

PR #846, opened 2022-08-29
Add `dataTestId` to Radio component

danoc:


egoens:
Looks like the Jest test is failing

danoc:
@egoens — This one is okay since it adds a new prop but keeps the old one around. The other one was problematic because it renamed the prop rather than deprecating the old one.

@lavelle — Can you take a look? 🙇🏻

lavelle (approved):

PR #847, opened 2022-09-01
Rearranges props to provide clearer order between props and component options

egoens:
The intention is to provide a better UX around finding available props and component information.


danoc (approved):
The build is failing for unrelated reasons that I need to look into. This looks great though.

egoens:
> The build is failing for unrelated reasons that I need to look into. This looks great though.

I'm assuming that means that we should hold off on merging until the build / test issues are resolved?

PR #849, opened 2022-10-05
fix(lodash): update lodash version

ben-codes-five:
Update lodash version to fix
prototype pollution vulnerability

Closes: FIXQ4-7076


danoc:
Nice work @ben-codes-five! Can you also update the chagnelogs of the other packages that changed? Looks like it was more than just Thumbprint React.

Also, the version numbers don't need to go in the changelog. See this doc for some tips:
https://github.com/thumbtack/thumbprint/blob/master/CONTRIBUTING.md#submitting-a-pull-request

zargold (dismissed):

PR #850, opened 2022-10-24
Adds october 2022 updates

egoens:
Ports the October 2022 newsletter into Thumbprint product updates.


danoc (approved):
Looks great! I think I just fixed the builds. This should be ready to merge in a few minutes.

PR #860, opened 2022-11-11
First ThumbprintUI documentation - Buttons!

dwroth:
Adds documentation for TPButton, TPRightIconButton, and TPIconButton (which has not yet been merged).

Also includes a change to the component navigation bar indicating that the old iOS Thumbprint documentation is no longer the most relevant.

image


www/src/components/thumbprint-components/component-header/platform-nav.jsx
```diff
@@ -32,21 +33,32 @@ const getPlatformDisplayName = path => {
         usage: 'Usage',
         react: 'React',
         scss: 'SCSS',
-        ios: 'iOS',
+        ios: 'iOS (Old)',
```
Abdul-Moiz, line 36: Nit: Should use UIKit instead of old. Reader should infer that automatically.

tt-pkaminski: +1, I'd probably not strikethrough the UIKit one since we'll still be using it for a long time to come.

dwroth: @egoens I didn't realize that banner thing existed.  I'll switch to using that instead.

www/src/components/thumbprint-components/component-header/index.jsx
```diff
@@ -2,6 +2,18 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import PlatformNav from './platform-nav';
 import ComponentPackageTable from './component-package-table';
+import Alert from '../../alert';
+
+const UIKitDeprecatedCopy = () => {
+    return (
+        <Alert type="warning" title="UIKit based Thumbprint is no longer being developed">
```
egoens, line 9: This is headed in the right direction, but this still feels a bit heavy. I'm wondering if we can come up with a pithier approach to this messaging and expand on what we have here a bit more in the November release notes.

dwroth: @egoens Sure.  Do you have any suggestions on what the copy should be?  

And @danoc do you have any suggestion on how to implement this without `window.location.pathname` which doesn't appear to be compatible with the documentation generator (if I'm reading that error message correctly).

www/src/components/thumbprint-components/component-header/index.jsx
```diff
@@ -2,6 +2,18 @@ import React from 'react';
...
+            As the iOS team is transitions from UIKit to SwiftUI backed view code, we are rebuilding
```
ChaoLiao, line 10: Nit: `is transitions` -> `transitions`?

www/src/components/thumbprint-components/component-header/index.jsx
egoens: @dwroth I'm currently floating this message with our content team, but I'm also making a proposal to them as well. Here is my suggestion for shortening this verbiage:

> Thumbprint is transitioning our codebase from UIKit to SwiftUI, and this documentation is no longer being supported. Please refer to the SwiftUI documentation when available.

dwroth: Oh, I like that a lot! 👍

egoens: Checked with CD for the compy here and got the 👍🏼

danoc (dismissed):
I think this is fixed. The problem is that the docs are built as a static website (using Node.js) 
but `window` only exists in the context of the browser. I added a small conditional (`typeof window !== 'undefined'`) so that the code doesn't error. The ideal solution would be such that this code works on both the server and client (browser), but I couldn't remember how to do that in Gatsby and it's not worth the effort IMO.

danoc (approved):

danoc (approved):

PR #862, opened 2022-11-28
Adds content design documentation

egoens:
This branch will add the content design documentation to Thumbprint. These changes include the addition to the Content Design in Overview section along with four sub-pages.

- Voice and tone
- Inclusion and accessibility
- Grammar and mechanics
- Surfaces and elements

The most notable change is how we are handling the Do/Don't component in the content pages. It's still be considered a potential global change, but for the time being, we are retaining the legacy X-out approach seen on the component pages.

### Legacy X-out version
image

#### Side-by-side
image

#### Table
image


www/src/components/dodont/index.tsx
```diff
@@ -5,32 +5,73 @@ import styles from './index.module.scss';
 
 interface PropTypes {
     children: React.ReactNode;
-    type: 'do' | 'dont';
+    type?: 'do' | 'dont';
+    coloredHeading?: boolean;
+    isTable?: boolean;
```
egoens, line 10: The table parameter was added to allow for direct comparisons of copy juxtapositioned. This also allows for far less redundancy that would otherwise overcrowd the page with visual do/don't callouts that could detract from the overall goal.

www/src/components/dodont/index.module.scss
```diff
@@ -1,19 +1,80 @@
...
+        &:last-of-type {
+            svg {
+                fill: $tp-color__red;
+            }
+            > div:last-of-type {
+                border-bottom: 2px solid $tp-color__red !important;
+            }
+        }
+    }
 }
```
danoc, line 80: Try to avoid nesting + element CSS selectors. Use classnames more, when possible.

lavelle (commented):
lgtm, will let dan approve

www/src/components/dodont/index.module.scss
egoens: @danoc this should be addressed with the most recent commit.

danoc (approved):
Nice work! Let's merge once the CI passes.

PR #870, opened 2023-01-06
Next.js #1: Set up Next.js docs skeleton site

danoc:
This is part of an effort to move the docs to Next.js. I'll be doing it in incremental commits throughout this week and next. The skeleton site was bootstrapped with `yarn create next-app`.

A lot of this code is copied from `www/*` in this repo. I decided to copy the files since it will make it easier and faster to do this incrementally. I had tried moving this site to Next.js nearly 2 years ago but the PR got too big and messy to be practical.

This is what it looks like:
Screenshot 2023-01-06 at 09 48 09


next/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "next",
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start"
```
egoens, line 8: Newb question but what's the difference between dev and start run scripts in nextjs?

egoens (approved):
Let the fun begin!

lavelle (approved):
didn't look to closely since it's mostly copying existing code

next/package.json
danoc: `dev` is for the local development environment and `start` is for running a production version of the site!

PR #873, opened 2023-01-10
Next.js PR 4: Implement the component overview page in Next.js

danoc:


next/components/thumbprint-components/overview/overview.jsx
```diff
@@ -0,0 +1,197 @@
+/* eslint-disable import/prefer-default-export */
```
lavelle, line 1: would be good to convert to TS but it can be a follow up

lavelle (approved):

PR #878, opened 2023-01-13
Next.js PR 9: Add component props table and more package info

danoc:
This uses `react-docgen` and `doctrine` to parse the React component files and create the props table. The stable version of `react-docgen` doesn't have types, so I added some. Also, I moved the MDX components (`H1`, `P`, `Table`, etc.) from `mdx/mdx.tsx` to `mdx/components.tsx` to avoid a cyclical import issue.


next/components/thumbprint-components/props-table/prop-type.jsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
```
lavelle, line 1: is this file copied from the old gatsby docs? do we still need to support proptypes if all components are written in TS now?

next/components/thumbprint-components/props-table/component-definition.ts
```diff
@@ -0,0 +1,24 @@
+export interface ComponentDefinition {
+    displayName: string;
+    description: string;
+    methods: unknown[];
```
lavelle, line 4: Is there a reason this is unknown?

next/components/thumbprint-components/props-table/component-definition.ts
```diff
@@ -0,0 +1,24 @@
...
+    props: Record<
+        string,
+        {
```
lavelle, line 7: Can this be a named type to make the nesting more readable?

next/components/mdx/mdx.tsx
```diff
@@ -289,11 +90,126 @@ export const MDXComponentPage = ({
...
+                                                            elementName="div"
+                                                        >
+                                                            <div className="b black-300">Type</div>
+                                                            <PropType
+                                                                type={prop.tsType.name}
+                                                                value={prop.tsType.elements}
+                                                            />
+                                                        </Text>
+                                                    )}
+                                                    {get(prop.defaultValue, 'value') && (
```
lavelle, line 185: prefer `?.` to `_.get` in TS files

lavelle (changes requested):

next/components/thumbprint-components/props-table/prop-type.jsx
danoc: Yeah this one is copied over.

next/components/mdx/mdx.tsx
danoc: I should've indicated this in the PR description, but this file was almost entirely copied over from the Gatsby code. I'd like to leave this as-is for now.

next/components/thumbprint-components/props-table/component-definition.ts
danoc: This only ever gets populated by class components, I think. I've updated this PR to expand on this type definition a bit.

next/components/thumbprint-components/props-table/component-definition.ts
danoc: Done

lavelle (approved):

PR #886, opened 2023-02-02
Add rest of content to the Atomic page

danoc:


lavelle (dismissed):
lgtm, didn't look too closely since i assume it's mostly copied from existing file

danoc:
Self approving this since the only change was a trivial merge conflict fix with the GH UI

PR #890, opened 2023-02-08
Next.js 15: Create a CodeExperimental component for Next.js docs

danoc:
I created this component since we're moving away from MDX. It allows developers to create live React/HTML examples without needing to write it all as a string.

I decided to call it CodeExperimental because I'm gradually adopting it and it's possible that we'll run into a issue with it that becomes a blocker. So far so good though. I'll rename this to `<Code>` and get rid of the existing `<Code>` component if this works well.


lavelle (changes requested):
I guess you did with with a find and replace, but does it make sense to be calling this component for every instance? Sometimes where it's a single word like `(c)` or `tpColorBlue` I feel like just using the HTML `<code>` tag would be better. We should only be calling this component when there's actual JSX inside it.

danoc:
> I guess you did with with a find and replace, but does it make sense to be calling this component for every instance? Sometimes where it's a single word like `(c)` or `tpColorBlue` I feel like just using the HTML `<code>` tag would be better. We should only be calling this component when there's actual JSX inside it.

My thinking was that I'd replicate the behavior of what MDX was doing under the hood by using those same components:

https://github.com/thumbtack/thumbprint/blob/2eb04d73acbc1124be4039062d8cbb080533edd2/www/src/components/mdx/index.tsx#L214-L230

We'd want people to use `<H2>` rather than `<h2>`, for example, because it adds [some helpful functionality](https://github.com/thumbtack/thumbprint/blob/master/www/src/components/mdx/index.tsx#L46-L56). I agree that it becomes iffier when the component is simply adding styles. Maybe the balance is that we should use plain-HTML tags for the ones that don't have any fancy functionality and then find a linter than can encourage usage of the components like `H2` that provide benefits only available in JS.

In the meantime I'd prefer to keep going down the direction I'm in since it guarantees that pages will look and behave like they did with MDX. Now that this is all TypeScript it'd be safe and easy to tweak this approach down the road.

lavelle (approved):
Alright, thanks for clarifying. I'll just approve for now to keep this project moving, I agree it makes more sense to just get something running on Next.js for now and then do any refactors we want to incrementally after we get the site itself live.

PR #892, opened 2023-02-13
Adds usage documentation for horizontal rule component

egoens:
Provides usage documentation for the horizontal rule component. This includes anatomy, sizing, and best practices such as do's and don'ts.


danoc (approved):
looks great!

PR #900, opened 2023-02-17
Prepare `/overview/` pages for Next.js launch

danoc:
This fixes small bugs I found in the Next.js docs so that the Next.js version of these pages looks and behaves the same as the Gatsby version.

It also removes `/overview/` from the Next.js config file, telling Next.js not to proxy that URL to the Gatsby version of the site.


next/components/layout/layout.tsx
```diff
@@ -128,20 +128,35 @@ export default function Layout({
...
-                                                                                href={
-                                                                                    sectionItem.href
-                                                                                }
-                                                                                key={
-                                                                                    sectionItem.href
-                                                                                }
-                                                                                level={3}
-                                                                            />
-                                                                        ),
+                                                                        sectionItem => {
```
lavelle, line 131: This is getting heavily nested to the point it would be better to extract into its own component

next/components/alert/alert.tsx
```diff
@@ -42,7 +42,9 @@ export default function Alert({ type, title, children }: PropTypes): JSX.Element
                     {title}
                 </Text>
             </div>
-            <div className={`black-300 ${styles.children}`}>{children}</div>
+            <Text className={`black-300 ${styles.children}`} size={2} elementName="div">
```
lavelle, line 45: Prefer classNames to string concat

lavelle (approved):
Leaving comments where I  think of them but approving since I know we want to power ahead with this project

next/components/layout/layout.tsx
danoc: Hah yeah, very fair. This component is unwieldy.

PR #905, opened 2023-02-21
Update the Next.js version of the DoDont components

danoc:
This is a direct copy of the Gatsby component into Next.js.

If keeping these components up to date becomes too much of a hassle we can look into creating a folder in this repo for shared components between both sites.


danoc:
@egoens — Can you take a look when you get a chance?

egoens (approved):
Looks great! LGTM!

PR #912, opened 2023-03-07
Add row-gap and col-gap classes to Atomic

lavelle:


packages/thumbprint-atomic/src/packages/flexbox.scss
```diff
@@ -205,6 +205,78 @@
     flex-shrink: 1 !important;
 }
 
+.col-gap0 {
```
danoc, line 208: This makes sense in its own file and documentation section, IMO, since these gap classes work for both flexbox or CSS grid. (I don't think there's much CSS grid use in our codebase, but it is production ready and should be added to Atomic eventually.)

danoc (commented):
One small comment. Thanks for doing this!

packages/thumbprint-atomic/src/packages/flexbox.scss
egoens: Silly question, but can we leverage scss for loops to generate incremental class structures like this?

lavelle: Moved it to its own file.

Yeah it's quite error prone doing it via copy-paste so I think using scss loops would be good. But yeah, we'd probably want to do it across the atomic codebase.

danoc (approved):
ty!

PR #918, opened 2023-04-03
Adds gap section to navigation in atomic page

egoens:
Small change to add a navigation option for the gap section in atomic.

image


danoc:
Thanks!! 🙇🏻

danoc (approved):

PR #919, opened 2023-04-04
Re-adds roadmap page (currently static)

egoens:
Adds roadmap page to provide more clarity around current projects that directly apply to the product.

### DIFF
--- a/www/src/components/container/index.tsx
+++ b/www/src/components/container/index.tsx
@@ -470,6 +470,12 @@ export default function Container({
                                 >
                                     <SideNavGroup level={2}>
                                         <SideNavLink
+                                            title="Roadmap"
+                                            level={2}
+                                            to="/updates/roadmap/"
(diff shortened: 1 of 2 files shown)


egoens:
@lavelle do you have an idea why the builds are falling? I've tried running `yarn build:docs` on master and I'm getting the same vague results. The only assumption I have is that it could be a missing `.env` for our Coda credentials, but that's only an assumption because it seems to be failing on the "/components/overview" page.

image

lavelle:
Hi @egoens, I'll refer this to @danoc. He's been moving our content from MDX to TSX files recently, which may be related to this new MDX file not working properly.

egoens:
Closing this for now until we find enough demand for having a publicly shared roadmap or until we can streamline with a Coda resource.

egoens:
Closing this for now until we find enough demand for having a publicly shared roadmap or until we can streamline with a Coda resource.

PR #922, opened 2023-04-26
Adds update information for April 2023 newsletter

egoens:
Adds April newsletter information to Thumbprint.


egoens:
@danoc Any thoughts on this PR? Looks like the netlify build process is failing on the `/comoponents/overview` page. Which is likely due to the CODA_API token issue that we've seen before.

danoc (approved):

PR #927, opened 2023-05-03
Refactor generic event props for Button and Link

lavelle:


lavelle:
This is ready for review, though I still need to figure out how to get react-docgen to show the full list of props even when using `extends` on the interface

egoens:
@lavelle after the lastest merge, it looks like there is a conflict on this PR now.

PR #928, opened 2023-06-08
Fix type and URL for Calendar disabledDays prop

lavelle:


danoc (approved):
Can you take take a look at the suggested changelog change? Otherwise LGTM. Feel free to self approve if you lose my approval from editing the changelog.

danoc (approved):

PR #944, opened 2023-07-14
Fixed types for the `Image` component

### LINKED ISSUE #940: <Image> does not support the `style` prop in TypeScript
There are examples in the [Thumbprint docs for the Image component](https://thumbprint.thumbtack.com/components/image/react/) that use the `style` prop to set the image width.

Image__React____Thumbprint

Currently, a TS error occurs if you try to add such a `style` prop to the `<Image>` component:

Fullscreen_7_10_23__1_27_PM

danoc:
We spread `props` on the `picture` element, but our types don't reflect that.

Fixes #940.

### DIFF
--- a/packages/thumbprint-react/components/Image/index.tsx
+++ b/packages/thumbprint-react/components/Image/index.tsx
@@ -22,7 +22,8 @@ type ImageSource = {
     srcSet: string;
 };
 
-export interface ImageProps {
+export interface ImageProps
+    extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement> {
     /**
      * If `sources` is provided, this image will be loaded by search engines and lazy-loaded for
      * users on browsers that don't support responsive images. If `sources` is not provided, this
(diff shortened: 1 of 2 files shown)


danoc:
@lavelle, @dcapo — PTAL

packages/thumbprint-react/components/Image/index.tsx
```diff
@@ -22,7 +22,8 @@ type ImageSource = {
     srcSet: string;
 };
 
-export interface ImageProps {
+export interface ImageProps
+    extends React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement> {
```
lavelle, line 26: Should this be HTMLPictureElement, instead of the generic HTMLElement?

danoc: I don't think so. When I hover on `<picture`, I get this type definition:
image

lavelle: Ah interesting, yeah I think that's technically a bug in the DOM typings, though in practice it doesn't matter because <picture> doesn't have any custom attributes, so it's equivalent to the generic interface. We can stick with this for consistency with the DOM typings

lavelle (approved):

PR #947, opened 2023-07-25
Upgrade react-day-picker to v8

BhoomikaSorathiya:
- Upgrade react-day-picker to v8.8.0
- Upgrade dependency date-fns v2.30.0


packages/thumbprint-react/components/Calendar/index.module.scss
```diff
@@ -5,76 +5,62 @@
     display: inline-block;
 }
 
-.root :global(.DayPicker) {
+.root :global(.rdp) {
```
BhoomikaSorathiya, line 8: Note: In react-day-picker v8, `DayPicker` class has been updated with  `rdp`.

packages/thumbprint-react/components/Calendar/index.tsx
```diff
@@ -170,39 +145,41 @@ export default function Calendar({
...
+                modifiersClassNames={modifiersClassNames}
+                onDayClick={(day): void => {
                     let newSelectedDays = normaliseValue(value);
 
                     if (allowMultiSelection) {
-                        if (selected) {
-                            const selectedIndex = findIndex(newSelectedDays, selectedDay =>
-                                DateUtils.isSameDay(selectedDay, day),
-                            );
+                        const selectedIndex = findIndex(newSelectedDays, selectedDay =>
```
BhoomikaSorathiya, line 179: Note: Only receiving `day` input from `onDayClick ` , so checking is selected day click or not manually.

packages/thumbprint-react/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## 14.3.6 - 2021-06-24
+
+### Changed
+
+-   [Patch] Upgrade `react-day-picker` to 8.8.0
+-   [Patch] Upgrade `date-fns` peer dependency to 2.30.0
```
danoc, line 8: It's technically not a peer dependency. Here's an overview of peer dependencies: https://nodejs.org/fa/blog/npm/peer-dependencies

```suggestion
-   [Patch] Upgrade `date-fns` to 2.30.0
```

packages/thumbprint-react/components/Calendar/index.module.scss
```diff
@@ -113,30 +99,37 @@
...
     width: $cell-side;
     min-width: 44px;
-    padding-top: $vertical-padding;
-    padding-bottom: $vertical-padding;
 }
 
-.root :global(.DayPicker-Day--disabled) {
-    color: $tp-color__gray;
-    cursor: default;
+.root :global(.rdp-cell:has(div)) {
```
danoc, line 104: `:has` doesn't work in Firefox yet. Is there a different way of doing this? https://caniuse.com/css-has

packages/thumbprint-react/components/Calendar/index.tsx
```diff
@@ -122,9 +97,9 @@ export interface CalendarProps {
     /**
      * A react-day-picker modifier for greater control over disabled days. Past selection is
      * disabled by default.
-     * See: https://react-day-picker-v7.netlify.app/docs/matching-days
+     * See: https://react-day-picker.js.org/api/types/Matcher
      */
-    disabledDays?: Modifier | Modifier[] | null;
+    disabledDays?: Matcher | Matcher[] | null;
```
danoc, line 102: My understanding is that this won't be a breaking change since only the name of the library's `disabledDays` prop has changed to `disabled`. I guess that means that `Modifier | Modifier[]` is equal to `Matcher | Matcher[]` in value?

packages/thumbprint-react/components/Calendar/index.tsx
```diff
@@ -170,39 +145,41 @@ export default function Calendar({
...
     }
 
     return (
         <div className={styles.root}>
             <DayPicker
-                disabledDays={disabledDays || undefined}
+                disabled={disabledDays || undefined}
                 fromMonth={get(disabledDays, 'before', null)}
                 toMonth={lastMonth}
-                month={month || selectedDays[0]}
```
danoc, line 189: Why was this prop removed? Seems like they still use it in v8: https://react-day-picker.js.org/basics/navigation

packages/thumbprint-react/components/Calendar/index.tsx
danoc: We used to return early if the day was `disabled`. Is that an issue?

packages/thumbprint-react/components/Calendar/test.tsx
```diff
@@ -314,10 +313,10 @@ describe('`onMonthChange` callback should be called on month change', () => {
 });
 
 describe('applying themes to Calendar day cells', () => {
-    const modifierPrefix = '.DayPicker-Day--';
+    const modifierPrefix = '.rdp-day--';
     const onChange = (): void => {};
     const onMonthChange = jest.fn();
-    const year = 2057;
```
danoc, line 320: Why does this have to change?

packages/thumbprint-react/components/Calendar/test.tsx
```diff
@@ -31,17 +26,21 @@ describe('calls `onChange` with the correct date', () => {
...
+            })
+            .at(0);
 
+        day15.simulate('click');
         expect(onChange).toHaveBeenCalledTimes(1);
         expect(onChange).toHaveBeenCalledWith([new Date('2057-09-15T07:00:00.000Z')]);
 
         // Click on the 22nd next.
-        const day22 = allDays.findWhere(n => n.prop('children') === 22);
+        const day22 = allDays.findWhere(n => n.text() === '22').at(0);
```
danoc, line 43: Do you know why `.at()` and `.hostNodes()` is suddenly needed?

packages/thumbprint-react/components/Calendar/index.tsx
```diff
@@ -33,13 +16,16 @@ function throwError(message: string): void {
 }
 
 export function normaliseValue(value: CalendarProps['value']): Date[] {
-    if (value === null) {
+    if (value === null || !value) {
```
lavelle, line 19: this can just be `if (!value)` because that will also handle the null case

packages/thumbprint-react/components/Calendar/index.tsx
```diff
@@ -33,13 +16,16 @@ function throwError(message: string): void {
...
         return [];
     }
 
-    const valueArr: DateIsh[] = castArray<DateIsh>(value);
+    if (!isArray(value)) {
+        const valueArr: string[] = castArray<string>(value.toString());
```
lavelle, line 24: What's the reason we're converting the value to a string in this line, and then back to a date in the next line?

lavelle (changes requested):
Dont have too much to add beyond Dan's comments

package.json
```diff
@@ -117,5 +117,8 @@
             "pre-commit": "pretty-quick --staged"
         }
     },
-    "packageManager": "yarn@3.2.1"
+    "packageManager": "yarn@3.2.1",
+    "dependencies": {
+        "date-fns": "^2.30.0"
```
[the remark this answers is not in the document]
BhoomikaSorathiya: Done!

packages/thumbprint-react/components/Calendar/index.tsx
BhoomikaSorathiya: In v7, for disabled date click, we were onDayClick() was being called with `disabled: true`. So we were manually return from this function for that case.  But,  In v8, for disabled date click, onDayClick() is not being called, So, no need to  return from it for `disabled` state.

And good point. I checked prop types again. We are receiving `selected` state in second param of onDayClick(), so removing manual selected day check and using `selected` param. (Apologies, I missed it initially.)

packages/thumbprint-react/components/Calendar/index.tsx
```diff
@@ -142,9 +117,9 @@ export interface CalendarProps {
     daysThemeStrikeout?: (date: Date) => boolean;
 }
 
-/**
- * Thin wrapper around `react-day-picker` that renders a calendar.
- */
+// /**
+//  * Thin wrapper around `react-day-picker` that renders a calendar.
+//  */
```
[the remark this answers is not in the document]
BhoomikaSorathiya: Done!

packages/thumbprint-react/CHANGELOG.md
BhoomikaSorathiya: Updated!

packages/thumbprint-react/components/Calendar/index.tsx
BhoomikaSorathiya: Yes, Both looks similar in type as mentioned below(Note: Name get changed little like RangeModifier to  DateRange, BeforeModifier to DateBefore):

V7: type Modifier = Date | RangeModifier | BeforeModifier | AfterModifier | BeforeAfterModifier | DaysOfWeekModifier | FunctionModifier | undefined;

V8: type Matcher = boolean | ((date: Date) => boolean) | Date | Date[] | DateRange | DateBefore | DateAfter | DateInterval | DayOfWeek;


Also, looks like new `boolean` matcher get added in v8.

packages/thumbprint-react/components/Calendar/index.tsx
BhoomikaSorathiya: Good point! I think I missed it from type check. Updated to `month `.👍

packages/thumbprint-react/components/Calendar/index.module.scss
BhoomikaSorathiya: I tried several options, but couldn't find anything relevant. Also posted on recat-day-picker [issues](https://github.com/gpbl/react-day-picker/issues/1854).

danoc: Oh, nice work! Sounds like they're going to fix it soon. Can you QA and see how looks in Firefox? Do we need to wait until they fix it, or is it usable?

BhoomikaSorathiya: I tested it in Firefox and it's look like this. Author has added it to milestone for react-day-picker@8.9.0.  So, not sure when it will get fixed. I tested functionality, everything looks good to me except minor cell border for outside day cells. @danoc What you suggest?  Should we can go with it or should we wait for fix in firefox? (Also testing in Safari & Chrome where it's LGTM.)

Screenshot 2023-08-02 at 5 12 28 PM

packages/thumbprint-react/components/Calendar/index.tsx
BhoomikaSorathiya: Good point. Updated!

packages/thumbprint-react/components/Calendar/test.tsx
BhoomikaSorathiya: I think because we are getting multiple nodes with same class. So we are using hostNodes() and at() to get single node of that.

Ex: `rdp-day--theme-strikeout` class applies to `<ForwardRef>` + `<button>`. So using hostNode(), we are getting single wrapper of that. 
Screenshot 2023-08-02 at 5 35 09 PM

packages/thumbprint-react/components/Calendar/index.tsx
BhoomikaSorathiya: Good catch! I think it's not needed!(I might played with it while fixing lint type issues). For non-array value, I can directly convert to array without type conversion to string and back to Date.

Updated!

packages/thumbprint-react/components/Calendar/index.tsx
```diff
@@ -33,13 +16,16 @@ function throwError(message: string): void {
...
 export function normaliseValue(value: CalendarProps['value']): Date[] {
-    if (value === null) {
+    if (!value) {
         return [];
     }
 
-    const valueArr: DateIsh[] = castArray<DateIsh>(value);
+    if (!isArray(value)) {
+        const valueArr: DateIsh[] = [value];
+        return map(valueArr, d => new Date(d));
```
lavelle, line 25: since this is guaranteed to be a single value, we can simplify these two lines to

`return [new Date(value)]`

that handles converting the value to a date, and wrapping in an array, in a single line

lavelle (dismissed):
LGTM, with one small nit, and assuming you've confirmed with Dan that the firefox CSS issue is non-blocking

packages/thumbprint-react/components/Calendar/index.tsx
BhoomikaSorathiya: Oppps! I missed it! Updated!

packages/thumbprint-react/components/Calendar/index.module.scss
BhoomikaSorathiya: @danoc  Thoughts?

zargold (approved):

packages/thumbprint-react/components/Calendar/index.module.scss
BhoomikaSorathiya: Taking it as a non-blocker and going to merge CR.

PR #953, opened 2023-09-19
Fix change log for thumbprint-react

BhoomikaSorathiya:


danoc:
Hi! I think @lavelle already fixed this? Maybe you two did this around the same time.

BhoomikaSorathiya:
Giles already fixed this in his release.

PR #963, opened 2023-10-26
Find 9054 Add rich tooltip support to existing tooltip component

BhoomikaSorathiya:
FIND-9054 Add rich tooltip support to existing tooltip component
- Add support for FormattedText support
- Add support for keyboard tab focus
- Update tooltip click behavior

### DIFF
--- a/packages/thumbprint-react/components/Tooltip/index.tsx
+++ b/packages/thumbprint-react/components/Tooltip/index.tsx
@@ -4,6 +4,7 @@ import assign from 'lodash/assign';
 import classNames from 'classnames';
 import { Manager, Reference, Popper, RefHandler } from 'react-popper';
 
+import { TextButton } from '../Button';
 import ConditionalPortal from '../../utils/ConditionalPortal';
 import useCloseOnEscape from '../../utils/use-close-on-escape';
 import canUseDOM from '../../utils/can-use-dom';
@@ -64,6 +65,21 @@ export interface TooltipProps {
      */
     text: string;
     /**
+     * Tooltip cta in form of text button/link for rich tooltip
+     */
+    cta?:
+        | {
+              type: 'link';
+              href: string;
+              text: string;
+              onClick?: () => void;
+          }
+        | {
+              type: 'button';
+              onClick: () => void;
+              text: string;
+          };
+    /**
      * Controls the look of the tooltip.
      */
     theme?: 'light' | 'dark';
@@ -92,6 +108,10 @@ export interface TooltipProps {
      * Adds a `z-index` to the tooltip. Before using this prop, try to use `container="inline"`.
      */
     zIndex?: number;
+    /**
+     * Persist tooltip and not hide on click
+     */
+    persistTooltipOnClick?: boolean;
 }
 
 export default function Tooltip({
@@ -102,55 +122,78 @@ export default function Tooltip({
     text,
     children,
     closeDelayLength = 200,
+    persistTooltipOnClick,
+    cta,
 }: TooltipProps): JSX.Element {
-    const [isOpen, setIsOpen] = useState<boolean>(false);
+    const [isOpen, setIsOpen] = useState<{
+        isOpen: boolean;
+        type?: 'click' | 'hover';
+    }>({
+        isOpen: false,
+    });
     const [openTimeout, setOpenTimeout] = useState<number | undefined>(undefined);
     const [closeTimeout, setCloseTimeout] = useState<number | undefined>(undefined);
 
-    const show = (): void => {
+    const show = (type: 'click' | 'hover'): void => {
         if (closeTimeout) {
             window.clearTimeout(closeTimeout);
         }
 
-        setIsOpen(true);
+        setIsOpen({
+            isOpen: true,
+            type,
+        });
     };
 
     const hide = (): void => {
-        setIsOpen(false);
+        setIsOpen({
+            isOpen: false,
+        });
     };
 
     const onFocus = (): void => {
         if (!doesWindowSupportTouch()) {
-            show();
+            show('hover');
         }
     };
 
     const onMouseEnter = (): void => {
         if (!doesWindowSupportTouch()) {
             // Trigger the tooltip to show after a small delay to prevent flickering.
-            setOpenTimeout(window.setTimeout(show, OPEN_TIMEOUT));
+            setOpenTimeout(window.setTimeout(() => show('hover'), OPEN_TIMEOUT));
         }
     };
 
     const onMouseLeave = (): void => {
-        // By default this adds a small delay before closing to improve the user experience.
-        setCloseTimeout(window.setTimeout(hide, closeDelayLength));
+        if (!(persistTooltipOnClick && isOpen.type === 'click')) {
+            // By default this adds a small delay before closing to improve the user experience.
+            setCloseTimeout(window.setTimeout(hide, closeDelayLength));
 
-        if (openTimeout) {
-            // When the mouse leaves we should clear any in-progress open timeouts, to prevent the
-            // tooltip from showing after the user is no longer hovering over the launcher.
-            clearTimeout(openTimeout);
+            if (openTimeout) {
+                // When the mouse leaves we should clear any in-progress open timeouts, to prevent the
+                // tooltip from showing after the user is no longer hovering over the launcher.
+                clearTimeout(openTimeout);
+            }
         }
     };
 
     const onClick = (): void => {
         if (doesWindowSupportTouch()) {
-            if (isOpen) {
+            if (isOpen.isOpen) {
                 hide();
             } else {
-                show();
+                show('click');
             }
+        } else if (persistTooltipOnClick) {
+            // Keep tooltip opened in case of desktop when persistTooltipOnClick is true
+            show('click');
         }
+    };
+
+    const onBlur = (): void => {
+        // By default this adds a small delay before closing which gives time
+        // to focus by keyboard tab inside tooltip content for rich tooltip.
+        setCloseTimeout(window.setTimeout(hide, closeDelayLength));
     };
 
     useCloseOnEscape(hide, canUseDOM);
@@ -167,7 +210,7 @@ export default function Tooltip({
                         onMouseEnter,
                         onFocus,
                         onMouseLeave,
-                        onBlur: hide,
+                        onBlur,
                         onClick,
                         ariaLabel: text,
                     })
@@ -175,7 +218,7 @@ export default function Tooltip({
             </Reference>
 
             <ConditionalPortal shouldDisplace={shouldDisplace}>
-                {canUseDOM && isOpen && (
+                {canUseDOM && isOpen.isOpen && (
                     <Popper
                         placement={position}
                         modifiers={{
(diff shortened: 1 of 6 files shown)


lavelle:
This is the JIRA ticket for this work. The video in that ticket explains more about the reasoning behind the change https://thumbtack.atlassian.net/browse/FIND-9054

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -153,7 +167,27 @@ export default function Tooltip({
...
+        // to focus by keyboard tab inside tooltip content for rich tooltip.
+        setCloseTimeout(window.setTimeout(hide, closeDelayLength));
+        setHasOpenedByOnClick(false);
+    };
+
+    const getAriaLabel = (): string => {
+        if (ariaLabel) {
+            return ariaLabel;
+        }
+        if (typeof text === 'string') {
```
michaeldrob, line 181: I think there's a utility function that can extract text from FormattedText. Might be good here, otherwise any rich text won't get proper treatment if ariaLabel isn't present

michaeldrob (commented):
I'm not sure how this works in thumbprint, but is a storybook story being added for this?

BhoomikaSorathiya:
> I'm not sure how this works in thumbprint, but is a storybook story being added for this?

Here is instance for rich tooltip: https://github.com/thumbtack/thumbprint/pull/963/commits/968847f3caabc969f61db54650c61241b44df5c6#diff-9c2f5c95a0f0ffa6d1d59466a3bb5e47acf65754ae098682770c7d8afcdb0e5dR44

packages/thumbprint-react/components/Tooltip/index.tsx
BhoomikaSorathiya: Yeah below function can be used to extract `ariaLabel` from `FormattedText`.

 _const ariaLabel = text.segments.map((segment) => segment.text).join('');_

But, text can have type of React.DOM / string. I am not sure how to extract **_ariaLabel_** from **_React.DOM_**.

michaeldrob: technically, you can get it with textContent by setting a ref against the DOM element, but I'm not sure it's worth going through this much trouble. Perhaps ariaLabel can be required if the input is React.DOM?

BhoomikaSorathiya: That's a good idea to use ariaLabel prop incase of  input is React.DOM. Let me add comment in ariaLabel prop types to be passed in that case. But, ariaLabel prop type will be non-required as it can be used for string inputs as well.

BhoomikaSorathiya: [Done](https://github.com/thumbtack/thumbprint/pull/963/commits/62fdb39a40d34c7c9b59a890da13d95183a7e4e9#diff-e873722ff811ec99525e10534f21cca386d7f8a89cf12299b5601ddb860969a7R100)

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -60,9 +60,9 @@ export interface TooltipProps {
      */
     children: (args: ChildrenPropTypes) => JSX.Element;
     /**
-     * Plain text that will appear within the tooltip. Links and formatted content are not allowed.
+     * ReactNode that will appear within the tooltip. Links and formatted content are allowed.
      */
-    text: string;
+    text: React.ReactNode | Text;
```
danoc, line 65: I chatted with @egoens and we'd rather have an API where the CTA is a separate prop. For example:

```
cta: {
  href: string;
  text: string;
} | {
  onClick: string;
  text: string;
}
```

This is better for both design consistency and accessibility since we can guarantee that the colors will be accessible.

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -213,13 +248,26 @@ export default function Tooltip({
...
-                                <WhenChildrenChange do={scheduleUpdate}>{text}</WhenChildrenChange>
+                                <WhenChildrenChange do={scheduleUpdate}>
+                                    <div
+                                        onFocus={(): void => {
+                                            if (closeTimeout) {
+                                                window.clearTimeout(closeTimeout);
+                                            }
+                                        }}
+                                        onBlur={hide}
+                                        className="pre-line"
```
danoc, line 266: We don't use Atomic classes in Thumbprint React due to some versioning issues it would introduce. You'll need to add to the Sass file.

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -102,10 +111,13 @@ export default function Tooltip({
     text,
     children,
     closeDelayLength = 200,
+    persistTooltipOnClick,
+    ariaLabel,
 }: TooltipProps): JSX.Element {
     const [isOpen, setIsOpen] = useState<boolean>(false);
     const [openTimeout, setOpenTimeout] = useState<number | undefined>(undefined);
     const [closeTimeout, setCloseTimeout] = useState<number | undefined>(undefined);
+    const [hasOpenedByOnClick, setHasOpenedByOnClick] = useState<boolean>(false);
```
danoc, line 120: Rather than making a separate `useState`, I think it'd be better to modify the `isOpen` state to be an object rather than a boolean:

```typescript
const [isOpen, setIsOpen] = useState<
    {
        isOpen: true,
        reason: 'click' | 'hover'
    } | {
        isOpen: false,
    }
>(false);
```

This guarantees that we avoid introducing a bug in the code where we're in an impossible state such as `isOpen = false` and `hasOpenedByOnClick = true`. It should also simplifies the code a bit since you won't have to "clean up" the value of `hasOpenedByOnClick` when `hide` or `setIsOpen`.

packages/thumbprint-react/components/Tooltip/index.tsx
BhoomikaSorathiya: Good to know! Updated!

packages/thumbprint-react/components/Tooltip/index.tsx
BhoomikaSorathiya: That's a good suggestion! Updated!

packages/thumbprint-react/components/Tooltip/index.tsx
BhoomikaSorathiya: I see! I am wondering what is `cta` needs to be place between somewhere in side text, not at the end. How does this work in that case?🤔

danoc: We'd re-design in that case to move the CTA to the end rather than in-between the text. Feel free to reach out to @egoens on Slack for design help!

BhoomikaSorathiya: Sure! Updated code to support cta!👍

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -102,58 +127,94 @@ export default function Tooltip({
...
+        if (ariaLabel) {
+            return ariaLabel;
+        }
+        if (typeof text === 'string') {
+            return text;
+        }
+        return 'Tooltip text';
+    };
+
+    useCloseOnEscape(() => {
```
danoc, line 215: This change can now be undone, I think.

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -102,58 +127,94 @@ export default function Tooltip({
...
     };
 
-    useCloseOnEscape(hide, canUseDOM);
+    const onBlur = (): void => {
+        // By default this adds a small delay before closing which gives time
+        // to focus by keyboard tab inside tooltip content for rich tooltip.
+        setCloseTimeout(window.setTimeout(hide, closeDelayLength));
+    };
+
+    const getAriaLabel = (): string => {
```
danoc, line 205: We can remove this function now since `text` will always be a string.

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -219,7 +280,37 @@ export default function Tooltip({
...
                                 https://github.com/thumbtack/thumbprint-archive/issues/1033 */}
-                                <WhenChildrenChange do={scheduleUpdate}>{text}</WhenChildrenChange>
+                                <WhenChildrenChange do={scheduleUpdate}>
+                                    <div
+                                        onFocus={(): void => {
+                                            if (closeTimeout) {
+                                                window.clearTimeout(closeTimeout);
+                                            }
+                                        }}
+                                        onBlur={hide}
```
danoc, line 290: Can you leave comments explaining why these are needed?

Is it not possible for that logic to live in the `<Popover>` component usage on L240?

egoens:
Here are some specs to help guide the approach. These specs can be found in the [Handoff -> V2 page](https://www.figma.com/file/T6C9iPBYOkbSO5XZzcXWGa/Tooltip?type=design&node-id=902-3484&mode=design&t=I4tlg1f3d3mLu6eW-11)

image

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -219,7 +280,37 @@ export default function Tooltip({
...
+                                            >
+                                                {cta.text}
+                                            </a>
+                                        )}
+
+                                        {cta && cta.type === 'button' && (
+                                            <TextButton onClick={cta.onClick}>
+                                                {cta.text}
+                                            </TextButton>
+                                        )}
```
danoc, line 311: Take a look at @egoens's latest comment. We should wrap in a `div` so that it becomes `display: block` rather tha appearing inline. Or wrap the `text` in a `p`.

packages/thumbprint-react/components/Tooltip/index.tsx
```diff
@@ -102,58 +127,94 @@ export default function Tooltip({
...
+        isOpen: boolean;
+        reason?: 'click' | 'hover';
+    }>({
+        isOpen: false,
+    });
     const [openTimeout, setOpenTimeout] = useState<number | undefined>(undefined);
     const [closeTimeout, setCloseTimeout] = useState<number | undefined>(undefined);
 
-    const show = (): void => {
+    const show = (reason: 'click' | 'hover'): void => {
```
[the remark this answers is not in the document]
BhoomikaSorathiya: Done

packages/thumbprint-react/components/Tooltip/index.tsx
BhoomikaSorathiya: Done

packages/thumbprint-react/components/Tooltip/index.tsx
BhoomikaSorathiya: Done

packages/thumbprint-react/components/Tooltip/index.tsx
BhoomikaSorathiya: Done! Wrapped in Div.

BhoomikaSorathiya:
> Here are some specs to help guide the approach. These specs can be found in the [Handoff -> V2 page](https://www.figma.com/file/T6C9iPBYOkbSO5XZzcXWGa/Tooltip?type=design&node-id=902-3484&mode=design&t=I4tlg1f3d3mLu6eW-11)
> 
> image

👍

danoc (dismissed):

danoc:
Nice work!

danoc (approved):
I will merge and release once CI passes

danoc (approved):

danoc:
@BhoomikaSorathiya — This was released as `@thumbtack/thumbprint-react` version `14.18.0`

PR #964, opened 2023-11-07
Remove "Was this page helpful?" form

danoc:
Unfortunately the form itself was not very helpful due to the amount of spam responses it received.

We can investigate adding this back someday if we can gate it to only employees.

### DIFF
--- a/www/package.json
+++ b/www/package.json
@@ -48,14 +48,12 @@
         "react-scroll-marker": "^0.2.1",
         "slugify": "^1.2.9",
         "url-join": "^4.0.1",
-        "uuid": "^3.4.0",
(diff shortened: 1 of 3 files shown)


egoens (approved):
Agreed. There isn't great value in this current experience. Maybe someday we'll find a more useful alternative to capturing feedback.

lavelle (approved):

egoens:
@danoc I just noticed this form is still present on a few pages.

Examples:
- https://thumbprint.design/components/action-sheet/ios/
- https://thumbprint.design/help/
- https://thumbprint.design/updates/notes/

PR #978, opened 2024-04-02
Fix types for Button component mouse and focus events

danoc:
These were supposed to have types but we never added them for some reason.


michaeldrob (approved):
LGTM

PR #981, opened 2024-06-08
Add Switch component

teckliew:
This add this the switch component that is used in settings and various parts of Thumbtack.
SCR-20240607-rauy
Screenshot 2024-06-06 at 4 49 59 PM

### DIFF
--- a/packages/thumbprint-react/components/Switch/index.module.scss
+++ b/packages/thumbprint-react/components/Switch/index.module.scss
@@ -1,0 +1,53 @@
+@import '~@thumbtack/thumbprint-tokens/dist/scss/_index';
+
+// Match the width/height of Thubmprint Checkbox
+$bar-height: 20px;
+$bar-height-large: 28px;
+$height-width-ratio: 1.8;
+$transition-duration: 0.1s;
+$transition-function: ease-in;
+
+.slideToggle {
+    width: $height-width-ratio * $bar-height;
+    height: $bar-height;
+    border-radius: $bar-height;
+    cursor: pointer;
+    transition: background-color $transition-duration $transition-function;
+
+    &Large {
+        width: $height-width-ratio * $bar-height-large;
+        height: $bar-height-large;
+    }
+
+    &Disabled {
+        user-select: none;
+        cursor: default;
+        opacity: 0.3;
+    }
+}
+
+.toggleHandle {
+    height: $bar-height - 2px; // Subtract 2px to account for top/bottom border of parent container
+    width: $bar-height - 2px; // Subtract 2px to account for left/right border of parent container
+    border-radius: 50%;
+    transition: transform $transition-duration $transition-function;
+    box-shadow: $tp-shadow__card;
+    box-sizing: border-box;
+
+    &Large {
+        height: $bar-height-large - 2px; // Subtract 2px to account for top/bottom border of parent container
+        width: $bar-height-large - 2px; // Subtract 2px to account for left/right border of parent container
+    }
+
+    .slideToggleChecked & {
+        transform: translate3d(($height-width-ratio - 1) * $bar-height, 0, 0);
+    }
+
+    .slideToggleLargeChecked & {
+        transform: translate3d(($height-width-ratio - 1) * $bar-height-large, 0, 0);
+    }
+
+    input:focus ~ & {
+        box-shadow: 0 0 1px 3px $tp-color__gray-300;
+    }
+}
--- a/packages/thumbprint-react/components/Switch/index.tsx
+++ b/packages/thumbprint-react/components/Switch/index.tsx
@@ -1,0 +1,117 @@
+import React from 'react';
+import classNames from 'classnames';
+import styles from './index.module.scss';
+
+export const SWITCH_COLOR = {
+    blue: 'blue',
+    green: 'green',
+};
+
+export const SWITCH_SIZES = {
+    small: 'small',
+    large: 'large',
+};
+
+export interface SwitchProps {
+    /**
+     * The `id` is added to the switch as an HTML attribute and passed to the `onChange`
+     * function.
+     */
+    id?: string;
+    /**
+     * Determines if the switch is checked.
+     */
+    isChecked?: boolean;
+    /**
+     * Function that runs when a switch value changes. It receives the new boolean value,
+     * and the provided `id` as such:
+     * `props.onChange(event.target.checked, props.id)`.
+     */
+    onChange: (isChecked: boolean, id: string) => void;
+    /**
+     * Disables the component.
+     */
+    isDisabled?: boolean;
+    /**
+     * Determines how the switch sizes. The default size is `small`.
+     */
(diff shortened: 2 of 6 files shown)


packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    accessibilityLabel,
+    color = SWITCH_COLOR.blue as keyof typeof SWITCH_COLOR,
+    size = SWITCH_SIZES.small as keyof typeof SWITCH_SIZES,
+}: SwitchProps): JSX.Element {
+    const bgColor = `bg-${color} b-${color}`;
+
+    const onToggle = (): void => {
+        onChange(!isChecked, id);
+    };
+    return (
```
mattybow, line 74: Do we want to allow switch to be used like checkbox in which you can pass it [children](https://thumbprint.design/components/checkbox/react/#section-multi-column-content)

packages/thumbprint-react/components/Switch/index.module.scss
```diff
@@ -0,0 +1,53 @@
+@import '~@thumbtack/thumbprint-tokens/dist/scss/_index';
+
+// Match the width/height of Thubmprint Checkbox
+$bar-height: 20px;
+$bar-height-large: 28px;
+$height-width-ratio: 1.8;
+$transition-duration: 0.1s;
+$transition-function: ease-in;
+
+.slideToggle {
```
mattybow, line 10: Should we change the name to match the component?

packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    isDisabled?: boolean;
+    /**
+     * Determines how the switch sizes. The default size is `small`.
+     */
+    size?: keyof typeof SWITCH_SIZES;
+    /**
+     * This adds an `aria-label` to the element.
+     * the importance of using labels](https://dequeuniversity.com/rules/axe/4.3/select-name).
+     */
+    accessibilityLabel?: string;
```
egoens, line 43: If this is optional, would we also need to support an `aria-labeledby` attribute? How do we support the presence of an accessible label outside of the component?

packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import classNames from 'classnames';
+import styles from './index.module.scss';
+
+export const SWITCH_COLOR = {
+    blue: 'blue',
+    green: 'green',
```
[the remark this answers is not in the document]
egoens: I'll address this in the design documentation, but the plan is to only support TT Blue here.

packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+};
+
+export const SWITCH_SIZES = {
+    small: 'small',
```
egoens, line 11: Should the default be `large`? Most use cases I've seen are large vs small.

packages/thumbprint-react/components/Switch/index.module.scss
```diff
@@ -0,0 +1,53 @@
+@import '~@thumbtack/thumbprint-tokens/dist/scss/_index';
+
+// Match the width/height of Thubmprint Checkbox
+$bar-height: 20px;
+$bar-height-large: 28px;
+$height-width-ratio: 1.8;
+$transition-duration: 0.1s;
```
egoens, line 7: Should we leverage the `$tp-duration__1` or `$tp-duration__2` token here instead? How did we come about selecting `0.1s` for the duration?

packages/thumbprint-react/components/Switch/index.module.scss
```diff
@@ -0,0 +1,53 @@
...
+$transition-function: ease-in;
```
egoens, line 8: Same here. Should we use `$tp-ease__in`?

packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+    isDisabled = false,
+    isChecked = false,
+    onChange,
+    dataTestId,
+    name,
+    accessibilityLabel,
+    color = SWITCH_COLOR.blue as keyof typeof SWITCH_COLOR,
+    size = SWITCH_SIZES.small as keyof typeof SWITCH_SIZES,
+}: SwitchProps): JSX.Element {
+    const bgColor = `bg-${color} b-${color}`;
```
danoc, line 69: We don't use Thumbprint Atomic classes within Thumbprint React because it would create a peer dependency between the two packages, complicating versioning.

packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+     */
+    dataTestId?: string;
+    /**
+     * Switches on a page with the same name will be grouped together when sent to the server.
+     */
+    name?: string;
+}
+
+export default function Switch({
+    id = '',
```
danoc, line 59: I don't think there's a benefit to defaulting to an empty string here. We don't do that for `name`, for example.

packages/thumbprint-react/components/Switch/index.tsx
danoc: @egoens's design doesn't include text, FWIW:
https://www.figma.com/design/OugkA9TKC19FC8eAvPZux1/Switch?node-id=1252-2886&t=jWXKAKoCNaZSHc5S-0

packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+        <div
+            className={classNames(styles.slideToggle, 'ba', {
+                [styles.slideToggleLarge]: size === SWITCH_SIZES.large,
+                [styles.slideToggleDisabled]: isDisabled,
+                [styles.slideToggleChecked]: isChecked,
+                [styles.slideToggleLargeChecked]: size === SWITCH_SIZES.large && isChecked,
+                [bgColor]: isChecked,
+                'bg-gray b-gray-300': !isChecked,
+            })}
+            onClick={isDisabled ? undefined : onToggle}
```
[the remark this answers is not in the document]
danoc: +1 — Doing it on the `input` is better for accessibility.

packages/thumbprint-react/components/Switch/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+                className="visually-hidden"
+                type="checkbox"
+                role="switch"
+                id={id}
+                data-testid={dataTestId}
+                disabled={isDisabled}
+                checked={isChecked}
+                name={name}
+                aria-checked={isChecked}
+                aria-disabled={isDisabled}
```
danoc, line 105: Are these needed if we have `type="checkbox"`? What did you use as a reference for the accessibility best practices?

PR #997, opened 2024-09-12
Make EntityAvatar round, deprecate all badges

danoc:
This is the same work that I've already done in the FE monorepo. I'm doing it here too so that the docs look up-to-date when we merge the code in the FE monorepo. This duplication won't be needed once we switch traffic to the new docs.


danoc:
Nevermind, I'm going to use Erik's PR for this.

